PR #12, opened 2021-11-26
Update list of allowed APIs to 4.3.0 level

vigejolla:


Develop/Apps/Harbour/Allowed_APIs/README.md
```diff
@@ -23,31 +23,20 @@ Your application can link against the following libraries:
   - libQt5Network.so.5
```
sledges, line 23: https://github.com/sailfishos/docs.sailfishos.org/blob/97d5ae1351b0c7b61ccd1914c890ecb8f9fce4af/Develop/Apps/Harbour/Allowed_APIs/README.md?plain=1#L10
Should this be also probably bumped?:)

vigejolla: 🤦

vigejolla: Fixed. Also fixed the Qt API part so that they are not grouped by Qt version

sledges (approved):

PR #13, opened 2021-11-30
Don't say Qt 5 is the latest major version of Qt

jpetrell:
Qt 6 came out end of the last year (though really 6.2 released few month back was the first mature version).


vigejolla (approved):
LGTM

PR #18, opened 2021-12-03
Update Qt logging categories for mKCal and KCalendarCore

dcaliste:
mKCal and KCalendarCore are now using standard Qt logging helper functions.


chriadam (approved):
LGTM, thanks.

PR #19, opened 2021-12-03
Propose to also use xdg-open to import calendar data

dcaliste:
One can also use xdg-open to import ICS data. It's simpler
and easier to remember than the pure DBus call.

### DIFF
--- a/Reference/Sailfish_OS_Cheat_Sheet/README.md
+++ b/Reference/Sailfish_OS_Cheat_Sheet/README.md
@@ -502,10 +502,12 @@ Import events from iCal
 devel-su -p icalconverter import calendar.ics
 ```
 
-Import events using Calendar import page
+Import events using Calendar import page (two possibilities)
 ```nosh
-dbus-send --print-reply --type=method_call --dest=com.jolla.calendar.ui /com/jolla/calendar/ui com.jolla.calendar.ui.importFile string:/home/<user>/calendar.ics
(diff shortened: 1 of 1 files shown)


Reference/Sailfish_OS_Cheat_Sheet/README.md
```diff
@@ -500,10 +500,12 @@ Import events from iCal
 devel-su -p icalconverter import calendar.ics
 ```
 
-Import events using Calendar import page
+Import events using Calendar import page (two possibilities)
 ```nosh
 dbus-send --print-reply --type=method_call --dest=com.jolla.calendar.ui /com/jolla/calendar/ui com.jolla.calendar.ui.importFile string:/home/<user>/calendar.ics
+xdg-open foo.ics
 ```
+This assumes that the ICS data are actually in a directory that is visible from inside the jail the calendar application is running into.
```
chriadam, line 508: This may need some more elaboration.  Maybe add: "For example, the home Documents folder" (assuming that one is accessible...)

chriadam:
One comment, aside from that LGTM.

Reference/Sailfish_OS_Cheat_Sheet/README.md
dcaliste: Done in the lastest version.

I've also change the DBus line to use `$HOME` and indicate that the file should be in a readable directory.

PR #22, opened 2021-12-15
Use /srv/sailfishos instead of /srv/mer

vigejolla:
JB#56724

### DIFF
--- a/Tools/Platform_SDK/Installation/README.md
+++ b/Tools/Platform_SDK/Installation/README.md
@@ -15,7 +15,7 @@ The Sailfish Platform SDK contains [Development Tools](/Tools/Development) like 
 
 If you're in a hurry then this should get you going (if it doesn't work then read the full instructions carefully!) :
 ```nosh
-export PLATFORM_SDK_ROOT=/srv/mer
+export PLATFORM_SDK_ROOT=/srv/sailfishos
 curl -k -O https://releases.sailfishos.org/sdk/installers/latest/Jolla-latest-SailfishOS_Platform_SDK_Chroot-i486.tar.bz2
 sudo mkdir -p $PLATFORM_SDK_ROOT/sdks/sfossdk
 sudo tar --numeric-owner -p -xjf Jolla-latest-SailfishOS_Platform_SDK_Chroot-i486.tar.bz2 -C $PLATFORM_SDK_ROOT/sdks/sfossdk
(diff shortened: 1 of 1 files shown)


Tools/Platform_SDK/Installation/README.md
```diff
@@ -65,7 +65,7 @@ The Platform SDK is provided as a rootfs tarball that contains essential tools f
 Filesystem requirements:
 
   - The Platform SDK can be installed to any location with enough space - we'll use /srv/ as per the [Linux FHS](http://www.pathname.com/fhs/pub/fhs-2.3.html#SRVDATAFORSERVICESPROVIDEDBYSYSTEM) (feel free to adapt the commands to use any other location).
-  - The installation path must contain an intermediate directory called 'sdks' which only has Platform SDKs inside. eg /srv/mer/sdks/sfossdk/...
+  - The installation path must contain an intermediate directory called 'sdks' which only has Platform SDKs inside. eg /srv/sailfishos/sdks/sfossdk/...
```
martyone, line 68: Maybe replace "sfossdk" with just "sdk" now as we already have "sailfishos" in the path?

There were also some complaints on using ~/SailfishOS instead of ~/SailfishSDK as the default installation path for THE SDK, so maybe use /srv/sailfishsdk as the prefix here?

vigejolla: Well, /srv/sailfishsdk/sdks/sdk looks a bit funny to me as well

vigejolla: Do we actually even need the "mer" or "sailfishos" directory there? Could we just have /srv/sdks/sfossdk?

vigejolla: answering myself: Yes we do, because targets and toolings are there. We probably don't want to have them directly under /srv.

Thaodan: > answering myself: Yes we do, because targets and toolings are there. We probably don't want to have them directly under /srv.

Also there's still the other sdk the ubuchroot until we build hal inside sfos.

vigejolla: Well, this page is not about ubuchroot, this is about platform sdk. And this PR is about using "sailfishos" (or "sailfishsdk" if that is somehow better) instead of "mer". I'd rather fix possible other issues in other PRs.

sledges (approved):

PR #32, opened 2022-02-04
Link instructions on setting up Docker in the FAQ

martyone:
The FAQ entry was there before and moved to the Installation page just recently. It is still referenced by the SDK Installer, so adding it back.


vigejolla (approved):
LGTM

PR #37, opened 2022-02-07
Fix build of Xperia 10 III

sledges:


llewelld (approved):
This did the trick for me in practice, so LGTM.

PR #40, opened 2022-02-10
Fix build of Xperia 10 II

sledges:


Thaodan:
Can you change the title to be more descriptive? E.g. Don't use ofono-ril-binder when building Xperia 10 II its not available in the current release.

sledges:
Changed.

PR #43, opened 2022-02-14
Instruct to use .bash_profile instead of .mersdk.profile

vigejolla:
There is no need to have a SDK specific initialisation file, as we can utilize the environment variable for SDK detection. JB#56751

### DIFF
--- a/Tools/Platform_SDK/Installation/README.md
+++ b/Tools/Platform_SDK/Installation/README.md
@@ -20,8 +20,10 @@ sudo mkdir -p $PLATFORM_SDK_ROOT/sdks/sfossdk
 sudo tar --numeric-owner -p -xjf Jolla-latest-SailfishOS_Platform_SDK_Chroot-i486.tar.bz2 -C $PLATFORM_SDK_ROOT/sdks/sfossdk
 echo "export PLATFORM_SDK_ROOT=$PLATFORM_SDK_ROOT" >> ~/.bashrc
 echo 'alias sfossdk=$PLATFORM_SDK_ROOT/sdks/sfossdk/sdk-chroot' >> ~/.bashrc; exec bash
-echo 'PS1="PlatformSDK $PS1"' > ~/.mersdk.profile
-echo '[ -d /etc/bash_completion.d ] && for i in /etc/bash_completion.d/*;do . $i;done' >> ~/.mersdk.profile
+echo 'if [[ $SAILFISH_SDK ]]; then' >> ~/.bash_profile
+echo '  PS1="PlatformSDK $PS1"' >> ~/.bash_profile
+echo '  [ -d /etc/bash_completion.d ] && for i in /etc/bash_completion.d/*;do . $i;done' >> ~/.bash_profile
+echo 'fi' >> ~/.bash_profile
 sfossdk
 ```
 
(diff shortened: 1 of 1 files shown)


Thaodan:
> There is no need to have a SDK specific initialisation file, as we can utilize the environment variable for SDK detection. JB#56751
Before this can/should be merged hadk should be updated.

Personally I don't like the idea polluting  SDK variables into my bashrc, those should go to a separate env or config file.

vigejolla:
> Before this can/should be merged hadk should be updated.

No, it doesn't have to be. Even if this documentation suggests to use .bashrc, the SDK still reads also .mersdk.profile, so HADK instructions keep on working. That being said, I agree, HADK should be changed as well, that just doesn't have to happen before this can be merged.

Thaodan:
> @Thaodan you can also still use `[[ $SAILFISH_SDK ]] && . /path/to/whatever.profile` in your .bashrc (as I do)

That doesn't really solve the underlying issue but working around it.

vigejolla:
Turns out my .profile, .bashrc etc. files did not quite have the content I thought they did, which caused quite a lot of confusion for me. When entering the the chroot, the bash indeed is login shell, and reads .bash_profile. The second bash it executes is not a login shell, but it will still read .bash_profile because that's set up in mer-bash-setup.

Tools/Platform_SDK/Installation/README.md
```diff
@@ -104,15 +106,17 @@ As mentioned, the Platform SDK is location independent so it uses the location o
...
-cat << EOF >> ~/.mersdk.profile
-PS1="PlatformSDK \$PS1"
-if [ -d /etc/bash_completion.d ]; then
-   for i in /etc/bash_completion.d/*;
-   do
+cat << EOF >> ~/.bash_profile
+if [[ $SAILFISH_SDK ]]; then
+  PS1="PlatformSDK \$PS1"
+  if [ -d /etc/bash_completion.d ]; then
+    for i in /etc/bash_completion.d/*;
```
martyone, line 115: Bash completion (as well as shell alias definitions, function definitions, shopt modifications etc.) is something that should go to .bashrc (or equivalent). .bash_profile is suitable for defining process environment as it is only read by the login shell. .bashrc is read by each interactive shell instance.

vigejolla: > .bashrc is read by each interactive shell instance.

That's what I though as well, but it turns out that's not true. If you have .bash_profile, .bashrc is not read when entering the sdk *unless* you have ". $HOME/.bashrc" in your .bash_profile. 

Note that e.g. Ubuntu has
```
if [ -n "$BASH_VERSION" ]; then
    # include .bashrc if it exists
    if [ -f "$HOME/.bashrc" ]; then
        . "$HOME/.bashrc"
    fi
fi
```
In default .profile, so .bashrc gets read if .bash_profile does *not* exist, but after creating .bash_profile it no longer gets read.

While I agree with you that bash_completion would make more sense in .bashrc, I have a feeling just moving it there is not quite enough. Should we also suggest adding that ". $HOME/.bashrc" thing to .bash_profile?

martyone: Yeah I should have written ".bashrc is read by other interactive shell instances" instead.

Yes, I believe reading .bashrc explicitly from .bash_profile is a must-have and I also believe every user account created a regular way will have it in the .bash_profile copied from `/etc/skel`.

Wouldn't some cleanup of the initialization code on sdk-setup side help?

It's all fun, isn't it? :)

vigejolla: > Yes, I believe reading .bashrc explicitly from .bash_profile is a must-have and I also believe every user account created a regular way will have it in the .bash_profile copied from `/etc/skel`.

You wish. Ubuntu does not have .bash_profile in /etc/skel. 

Should I just add adding 
```
if [ -f "$HOME/.bashrc" ]; then
  . "$HOME/.bashrc"
fi
```
to .bash_profile to the instructions?

> Wouldn't some cleanup of the initialization code on sdk-setup side help?

Quite likely. It's just that it's so convoluted that I don't even know where to start. Well, maybe I could start by simply adding
```
[[ -e ~/.bashrc ]] && . ~/.bashrc
```
to mer-bash-setup, before reading .bash_profile? Not exactly cleanup, but it would help with this case. What do you think?

> It's all fun, isn't it? :)

Depends on your definition of fun :)

martyone: IIRC we only need .bashrc in our instructions for loading bash completion. But if the regular bash rc files are used, then we do not need to do anything about that, so that could be simply dropped and after that the question whether .bashrc is read or not is not relevant anymore to us.

On sdk-setup, I would start by ensuring that sdk-chroot behaves equally as bash. Now it reads .bash_profile with the help of --init-file, which is meant to be used to replace .bashrc, so it looks rather incorrect. So would it be possible to simply let bash do what it would do by default?

martyone:
Rebased

martyone:
Updated after changes on sdk side, polished. Bash completion is going to be loaded automatically.

mkosola (approved):

rainemak (approved):

vigejolla:
rebased once again

PR #45, opened 2022-02-17
HADK changes after introducing `mb2 --output-dir`

sledges:


Develop/HW_Adaptation/Sailfish_X_Xperia_Android_10_Build_and_Flash/README.md
```diff
@@ -38,6 +38,8 @@ export HABUILD_DEVICE=pdx201
 export FAMILY=seine
 export ANDROID_VERSION_MAJOR=10
 export HAVERSION="sony-aosp-"$ANDROID_VERSION_MAJOR
+# Uncomment the next line to conveniently build all RPMs in local repo:
+#alias mb2="mb2 --output-dir $ANDROID_ROOT/droid-local-repo/$DEVICE"
```
martyone, line 42: Now I realized we might do better with late substitution, quoting and some error checking.

```suggestion
#alias mb2='mb2 --output-dir "${ANDROID_ROOT?}/droid-local-repo/${DEVICE?}"'
```

sledges: Thanks that's great, alias with late substitution will be added to HADK pdf, so we don't have to repeat it here per-port. Commit removed from PR \o/

martyone (approved):

PR #52, opened 2022-03-25
Move logging instructions from Browser page

rainemak:
This commit moves logging instructions from Browser page to here.


llewelld (approved):
Nice changes, LGTM.

PR #55, opened 2022-03-27
Set correct indention

pherjung:
download is described as method, without indentation it's a function.

### DIFF
--- a/Develop/Apps/Tutorials/Creating_an_application_in_Python/README.md
+++ b/Develop/Apps/Tutorials/Creating_an_application_in_Python/README.md
@@ -72,17 +72,17 @@ This function takes five seconds to run. Twice a second it will send a signal wh
 ```python
 class Downloader:
     def __init__(self):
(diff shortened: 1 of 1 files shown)


martyone (approved):

Develop/Apps/Tutorials/Creating_an_application_in_Python/README.md
```diff
@@ -77,12 +77,12 @@ class Downloader:
     self.bgthread = threading.Thread()
     self.bgthread.start()
```
llewelld, line 78: It looks like lines 75-78 should be inside the `__init__` method and so also need additional indentation. Would it be possible to fix this at the same time please?

llewelld (changes requested):
Thanks for the PR @pherjung. The changes you've suggested all look good, but I think it would make sense to fix the incorrect indentation in the few lines above at the same time, if that would be okay?

llewelld (approved):
Great, many thanks! This looks good to me now.

PR #58, opened 2022-03-31
Remove ChangeLog from Sailfsh X Android build and flash, update OEM blobs mentioned

Thaodan:
- Remove ChangeLog imported from the Wiki, the git history is much better for this
- Update OEM blob file for Android 11 based AOSP flash instructions, V9a based binaries are out and already used everywhere else.


llewelld (approved):
The change looks good to me.

As a side note though, it's not clear to me whether this is the changelog for the documentation (in which case, shouldn't the git history be enough?) or covering all repos related to the 10 III adaptation.

Thaodan:
> The change looks good to me.
> 
> As a side note though, it's not clear to me whether this is the changelog for the documentation (in which case, shouldn't the git history be enough?) or covering all repos related to the 10 III adaptation.

Removed the changelog instead.

llewelld (approved):
LGTM!

Thaodan:
Fixed a small typo.

llewelld (approved):
Still LGTM!

PR #64, opened 2022-04-11
Update Export local contacts to vCard section

nephros:
See also: https://forum.sailfishos.org/t/exporting-contacts-people-with-vcardconverter-doesnt-work/11132/2

### DIFF
--- a/Reference/Sailfish_OS_Cheat_Sheet/README.md
+++ b/Reference/Sailfish_OS_Cheat_Sheet/README.md
@@ -545,6 +545,16 @@ Export local contacts to vCard
 ```nosh
 devel-su -p vcardconverter --export contacts.vcf
 ```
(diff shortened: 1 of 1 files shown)


Reference/Sailfish_OS_Cheat_Sheet/README.md
```diff
@@ -545,6 +545,16 @@ Export local contacts to vCard
 ```nosh
 devel-su -p vcardconverter --export contacts.vcf
```
Ric9k-TMO-FSO, line 546: This will not work if done as root. Only execute this command as normal user with devel-su.

rainemak: Command is not root rather privileged user. This is exactly as it should be as privileged rights are needed for opening contacts database.

PR #67, opened 2022-04-20
Cleanup reporting issues

rainemak:


Thaodan:
I'm wondering if the commit message should reflect the change, e.g. in this case remove the mentions that issues can be reported on Github if its an OSS component.

PR #69, opened 2022-04-20
Add doc site previewing instructions for Fedora

rainemak:


Thaodan (changes requested):
Please separate the setup for jekyll.
The instruction are the same after installing dependencies, no need to duplicate them.

rainemak:
> Please separate the setup for jekyll. The instruction are the same after installing dependencies, no need to duplicate them.
> 
> I would first separate the dependency install instruction from the bundler and then jekyll setup. After that add a commit to describe installing the dependencies for Fedora.

It's really matter of taste. Having one set of instructions is simpler to follow when you're new to this. Updated like you proposed.

rainemak:
Rebased

Thaodan (approved):
LGTM. 

I'm not sure why gem install on Ubuntu needs sudo, the section for instructions for that should be the same.

sledges:
https://stackoverflow.com/questions/37720892/you-dont-have-write-permissions-for-the-var-lib-gems-2-3-0-directory :)

PR #76, opened 2022-05-05
Add Sailfish Platform SDK to the Tools navigation bar

rainemak:
This PR contains two commits. First one adds "Sailfish Platform SDK" to the side navigation bar of the Tools. Second commit renames "Platform SDK" to "Sailfish Platform SDK".

Before this change "Platform SDK" was well hidden behind HADK section of the Tools.

### DIFF
--- a/Tools/Platform_SDK/Installation/README.md
+++ b/Tools/Platform_SDK/Installation/README.md
@@ -6,7 +6,7 @@ layout: default
 nav_order: 100
 ---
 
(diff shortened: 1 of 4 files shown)


rainemak:
> The [Sailfish Platform SDK](/Tools/Platform_SDK) provides a similar integrated development environment to the Sailfish SDK, but targetted at developing the Sailfish OS operating system components rather than Sailfish OS applications.

This is good suggesting. Will amend this there.

PR #78, opened 2022-05-10
Add "View page history" link to every page

llewelld:
Adds a link to the bottom of each page, next to the "Edit this page on
GitHub" link, which links through to the github commit history. Can be
controlled using the _config.yml file.

See JB#57952

### DIFF
--- a/_config.yml
+++ b/_config.yml
@@ -116,6 +116,8 @@ gh_edit_repository: "https://github.com/sailfishos/docs.sailfishos.org" # the gi
 gh_edit_branch: "master" # the branch that your docs is served from
 # gh_edit_source: docs # the source that your files originate from
 gh_edit_view_mode: "tree" # "tree" or "edit" if you want the user to jump into the editor immediately
+gh_edit_history_link: true # show or hide commit history link
+gh_edit_history_text: "View page history"
 
 # Color scheme currently only supports "dark", "light"/nil (default), or a custom scheme that you define
 color_scheme: sfos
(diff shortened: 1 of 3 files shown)


_config.yml
```diff
@@ -116,6 +116,8 @@ gh_edit_repository: "https://github.com/sailfishos/docs.sailfishos.org" # the gi
 gh_edit_branch: "master" # the branch that your docs is served from
 # gh_edit_source: docs # the source that your files originate from
 gh_edit_view_mode: "tree" # "tree" or "edit" if you want the user to jump into the editor immediately
+gh_edit_history_link: true # show or hide commit history link
+gh_edit_history_text: "Changelog"
```
rainemak, line 120: Should this be "Change history"? I find "Changelog" a bit techy.

llewelld: Agreed; "Change history" is clearer. I've changed it.

Thaodan:
Shoudn't it be view history instead of change history?

rainemak:
> Shoudn't it be view history instead of change history?

Not necessarily. "Change history" can be treated as action toward history or like here "see how it has changed over time".

rainemak:
Once agreed, please update also commit title and PR title.

rainemak (approved):
It's already while back when I tested this last time but change looks still the same. Hence, approving.

llewelld:
I agree with both: "Change history" was intended to mean "View the *change history*", but since it was clearly a bit unclear I've changed it to "View page history", which I think is clearer.

I've made this change, updated the commit title and the PR title, and tested it locally (it works as before). The changes are minor, so I'll go ahead and merge.

PR #79, opened 2022-05-10
Clean out unavailable packages from  Xperia 10 III pattern.

mkosola:


sledges (approved):

Thaodan:
Can you add why the need to be removed so they user understands the reason.

mkosola:
> Can you add why the need to be removed so they user understands the reason.

Done

PR #81, opened 2022-05-10
Don't remove sailfish-devicelock-fpd from droid-configs on local build

Thaodan:
The dependencies for it namely sailfish-fpd-slave binder are now in
hw-common making sailfish-fpd also work for local builds.


sledges (approved):

Thaodan:
Rebased and merged.

PR #82, opened 2022-05-11
Add a Known Issues section

llewelld:
Introduces a Known Issues subsection to the Releases section, for
documenting the known issues in each release.

Pre-populates the subsection with the Known Issues taken from the
Release Notes of the last five releases (all the Sailfish 4 releases up
to this point).

### DIFF
--- a/Releases/Known_Issues/Koli/README.md
+++ b/Releases/Known_Issues/Koli/README.md
@@ -1,0 +1,76 @@
+---
+title: Sailfish OS 4.0.1.48 Known Issues
+permalink: Releases/Known_Issues/Koli/
+parent: Known Issues
+grand_parent: Releases
+layout: default
+nav_order: 500
+---
+
+This content can also be found in the [forum Release Notes](https://forum.sailfishos.org/t/release-notes-koli-4-0-1/4542#known-issues-generic-42).
+
+## Generic
+
+- *Assisted location service of Mozilla* has stopped working from 2020-03-01. This tends to slow down finding the position of the phone. We are looking for alternative ways to make it work faster. Unfortunately, this problem persists. It seems to affect Xperia XA2 worst.
+- Sailfish can connect *Bluetooth Low Energy devices*. However, Android apps cannot use peripheral devices via Bluetooth - other than for playback of the sound. Therefore smartwatches, for instance, cannot be controlled from Android apps, unfortunately.
+- *Signing in to Dropbox* works via Google only. So, after triggering the sign-in at "Settings > Accounts > Dropbox" and the browser page dropbox.com appears, tap "Sign in with Google" (if you have a Google account).
+- *Importing contacts wizard* fails to create a Google account in "Settings > Apps > People". The same happens if one tries to sign in to a Google account from Email. However, signing in to Google from "Settings > Accounts" does work correctly, after which the contacts will be downloaded to the device (if you allowed this in the account settings) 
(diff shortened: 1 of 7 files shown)


rainemak:
This would repeat the same "problem" that we have with changelog in forum. Known issues that have been there since release x are repeated. I see following factors at least: release where known issue appeared, device (sw component), release when known issues got fixed.

I was thinking more like a table like approach but grouping actually work as well. 

What if the high-level items in the left hand side navigation bar would be:

Generic
Android App Support 10 - Xperia 10 II, Xperia 10, and Xperia XA2
Xperia 10 II
Xperia 10
Xperia XA2
Xperia X
Jolla Tablet
Jolla C
Gemini PDA

Then each issue could have when seen first time and when fixed. For example

Under Xperia 10:


| Known issue          | Since release      | Fixed release |
| --------------------      | --------------------    | ---------------     |
| Top edge swipes in landscape orientation may not work very well | Sailfish OS 4.1.0.24 |  | 

This way we'd not need to repeat anything. Just mark known issues as fixed to the table. What do you think?

sledges:
If you decide to go the table route, the best option will be to rewrite it in HTML (can still use Markdown in its cells).

Such a big table in Markdown would introduce a few hurdles, such as cannot use [codeblocks within](https://stackoverflow.com/questions/28508141/code-block-inside-table-row-in-markdown), and readability/editability (try to edit my comment to see its source).

| Known issue          | Since release      | Fixed release |
| --------------------      | --------------------    | ---------------     |
| *Assisted location service of Mozilla* has stopped working from 2020-03-01. This tends to slow down finding the position of the phone. We are looking for alternative ways to make it work faster. Unfortunately, this problem persists. It seems to affect Xperia XA2 worst. |
| Sailfish can connect *Bluetooth Low Energy devices*. However, Android apps cannot use peripheral devices via Bluetooth - other than for playback of the sound. Therefore smartwatches, for instance, cannot be controlled from Android apps, unfortunately. | Sailfish OS 4.1.0.24 | |
| *Signing in to Dropbox* works via Google only. So, after triggering the sign-in at "Settings > Accounts > Dropbox" and the browser page dropbox.com appears, tap "Sign in with Google" (if you have a Google account). | Sailfish OS 4.1.0.24 | |
| *Importing contacts wizard* fails to create a Google account in "Settings > Apps > People". The same happens if one tries to sign in to a Google account from Email. However, signing in to Google from "Settings > Accounts" does work correctly, after which the contacts will be downloaded to the device (if you allowed this in the account settings). | Sailfish OS 4.1.0.24 | |
| An invalid warning and request to *uninstall all system packages* will appear if the previous OS version 3.4.0 is flashed to a device without installing the default Sailfish apps, followed by the download of the current OS version 4.0.1. This warning looks scary but it is just a warning. It should be okay to continue downloading and installing the OS update in the normal manner, i.e., you can ignore the warning. To avoid the warning, enabling the developer mode and running <code> devel-su; pkcon refresh</code> before downloading 4.0.1 may prevent the warning from appearing. | Sailfish OS 4.1.0.24 | |
| *Call recorder* does not save the conversation to the phone storage. This can be fixed with the following commands (and in the next OS release) - make sure you type the whole long command on the 2nd row:
```
        devel-su
        echo -e '\nmkdir     ${PRIVILEGED}/Phone\nprivileged-data Phone' >> /etc/sailjail/permissions/Phone.permission 
        systemctl-user restart voicecall-ui-prestart
``` | Sailfish OS 4.1.0.24 | |

jovirkku:
I'll try and test the table approach that Raine proposed. It would certainly reduce repetition.

llewelld:
This has now been superseded by @jovirkku's PR #112, so I'm closing this.

PR #83, opened 2022-05-11
Add warning regarding user added adaptation-common

rainemak:
Add warning regarding user added adaptation-common. Users should
never add adaptation-common themselves rather it should be the one
coming from the global ssu settings.

See JB#57866

### DIFF
--- a/Reference/Sailfish_OS_Cheat_Sheet/README.md
+++ b/Reference/Sailfish_OS_Cheat_Sheet/README.md
@@ -404,8 +404,22 @@ ssu lr                  # list
 ssu updaterepos
 pkcon refresh
 pkcon update
-```    
-    
+```
+
+Never override global repositories such as *adaptation-common* as that will mess up your device when upgrading it. Before upgrade make sure *Enabled repositories (global)* are not listed in *Enabled repositories (user)* as follows:
+
+```nosh
+devel-su ssu lr
+```
+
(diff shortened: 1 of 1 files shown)


Reference/Sailfish_OS_Cheat_Sheet/README.md
```diff
@@ -404,8 +404,22 @@ ssu lr                  # list
 ssu updaterepos
 pkcon refresh
 pkcon update
-```    
-    
+```
+
+Never override *adaptation-common* repository as that will messup your device when upgrading it. Before upgrade make sure *adaptation-common* repository is not listed in *Enabled repositories (user):* as follows:
```
saukko, line 409: This is not only adaptation-common but any global repository, right?

sledges, line 409: nit:
```suggestion
Never override *adaptation-common* repository as that will mess up your device when upgrading it. Before upgrade make sure *adaptation-common* repository is not listed in *Enabled repositories (user):* as follows:
```

rainemak: Thanks.

rainemak: I guess it's better to rephrase that way just to be on the safe side.

rainemak: Adjusted

mkosola (approved):

PR #89, opened 2022-06-21
Remove developer mailing list reference

rainemak:
Focus should on the Sailfish OS Forum instead.


llewelld (approved):
Sensible change, LGTM.

PR #95, opened 2022-08-07
Add export/import for call logs and messages

neomilium:
Some commands that I expected to find when I read this cheat sheet.

BTW, thanks for it!

### DIFF
--- a/Reference/Sailfish_OS_Cheat_Sheet/README.md
+++ b/Reference/Sailfish_OS_Cheat_Sheet/README.md
@@ -511,6 +511,16 @@ Install commhistory-tool if not already installed.
 devel-su -p pkcon install libcommhistory-qt5-tools
 ```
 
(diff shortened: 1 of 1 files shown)


Reference/Sailfish_OS_Cheat_Sheet/README.md
```diff
@@ -511,6 +511,16 @@ Install commhistory-tool if not already installed.
 devel-su -p pkcon install libcommhistory-qt5-tools
 ```
 
+Export call logs
```
rainemak, line 514: Just to make it super clear worth adding ", run as user" and same for the other.

neomilium: Commit updated.
Thanks.

rainemak (approved):
Thank you @neomilium for your contribution.

PR #97, opened 2022-08-09
Replace DBusActivatable with X-DBusActivatable. JB#57377

Tomin1:
We support both versions because it's non-standard to use
DBusActivatable without formatting the file name according to reverse
domain name rules.

See also https://github.com/sailfishos/lipstick/pull/27.

### DIFF
--- a/Reference/Core_Areas_and_APIs/Apps_and_MW/Lipstick/Launchers/README.md
+++ b/Reference/Core_Areas_and_APIs/Apps_and_MW/Lipstick/Launchers/README.md
@@ -50,14 +50,20 @@ showing a window too early.
 
 Sailfish OS 4.5.0 and later support using [_org.freedesktop.Application_ D-Bus
 interface](https://specifications.freedesktop.org/desktop-entry-spec/desktop-entry-spec-latest.html#dbus)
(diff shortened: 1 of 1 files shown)


Reference/Core_Areas_and_APIs/Apps_and_MW/Lipstick/Launchers/README.md
```diff
@@ -50,12 +50,13 @@ showing a window too early.
 
 Sailfish OS 4.5.0 and later support using [_org.freedesktop.Application_ D-Bus
 interface](https://specifications.freedesktop.org/desktop-entry-spec/desktop-entry-spec-latest.html#dbus)
-for opening files and activating windows. By specifying _DBusActivatable=true_ in _Desktop Entry_
+for opening files and activating windows. By specifying _X-DBusActivatable=true_ in _Desktop Entry_
```
Tomin1, line 53: I'm not sure if I should be a bit more verbose about this "X-" part here, but feel free to comment.

Thaodan: Maybe add the reasoning for why this was done/context and thous also the reason/necessary change needed to use the standard version.

Thaodan (approved):
LGTM

PR #104, opened 2022-08-18
Drop Sailfish SDK links from the Platform SDK doc

rainemak:
We do have two SDK. Let's keep it that way:
- Sailfish Platform SDK is primarily used by platform developers and documented option for hardware adaptation development
- Sailfish SDK is more for application developers

This should make it clearer for new hardware adaptation developers which one to use.

See JB#58565

### DIFF
--- a/Tools/Platform_SDK/README.md
+++ b/Tools/Platform_SDK/README.md
@@ -6,11 +6,9 @@ layout: default
 nav_order: 200
 ---
 
-**Attention: Platform development is now more convenient with the regular [Sailfish SDK](/Develop/Platform) through its command line frontend 'sfdk', available on all Linux, macOS and Windows. The Linux-only chroot based Sailfish Platform SDK remains available for special applications. It also remains the default (documented) option for [Hardware Adaptation Development](/Tools/Hardware_Adaptation_Development_Kit), although the regular Sailfish SDK can be used for that purpose as well.**
-
 # Overview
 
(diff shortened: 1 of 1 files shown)


vigejolla:
> We do have two SDK. Let's keep it that way:
> 
>     * Sailfish Platform SDK is primarily used by platform developers and documented option for hardware adaptation development
>     * Sailfish SDK is more for application developers

This is not true. The Sailfish SDK is both for application developers and platform developers. It can be used for hardware adaptation development as well, but like you wrote, it's not the documented option for that purpose.

Thaodan:
> > We do have two SDK. Let's keep it that way:
> > ```
> > * Sailfish Platform SDK is primarily used by platform developers and documented option for hardware adaptation development
> > * Sailfish SDK is more for application developers
> > ```
> 
> This is not true. The Sailfish SDK is both for application developers and platform developers. It can be used for hardware adaptation development as well, but like you wrote, it's not the documented option for that purpose.

Both work but the old text labels the platform sdk as worse of the two while it is more about preference.
In case of platform development or hardware adaptations the Sailfish SDK without the platform uses the same system as the chroot but with the sfdk wrapper with all it's advantages and disadvantages (e.g. Docker, path conversion etc.).

rainemak:
Superseded by https://github.com/sailfishos/docs.sailfishos.org/pull/152

PR #110, opened 2022-09-08
Remove WebKit reference from email rendering

rainemak:
Nowadays all (text/plain and text/html) emails are rendered with Sailfish WebView and QtWebKit has been deprecated.

See https://forum.sailfishos.org/t/deprecation-notice-silicawebview-qtwebkit


vigejolla (approved):
LGTM

PR #111, opened 2022-09-13
Remove WebKit reference from email rendering part2

rainemak:


vigejolla (approved):
Approved, even though it's worth considering if we should link to API docs instead of source repo

PR #115, opened 2022-09-21
Add building a package with mb2

vigejolla:


Tools/Platform_SDK/Building_Packages/README.md
```diff
@@ -0,0 +1,74 @@
...
+│ └─SailfishOS-4.4.0.58-armv7hl.default  (snapshot)
+└─SailfishOS-4.4.0.58-i486
+  └─SailfishOS-4.4.0.58-i486.default  (snapshot)
+```
+
+Each build target lets you build software for a specific operating
+system version running on a specific hardware. Each target is listed
+under one tooling supporting the given OS version. Under each build
+target are listed existing snapshots of the target. So, in our
+example, we have one installed one tooling (SailfishOS-4.4.0.58) and
```
dcaliste, line 34: There is a 'one' that seems to be too much here, do yu mean "we have installed one tooling" ?

sledges:
Please do not hard-wrap the text, otherwise it will be tough to re-wrap it every time it's modified and to read its diff too:)

vigejolla:
> Please do not hard-wrap the text, otherwise it will be tough to re-wrap it every time it's modified and to read its diff too:)

Really? M-x fill-paragraph is hard?

vigejolla:
> Please do not hard-wrap the text, otherwise it will be tough to re-wrap it every time it's modified and to read its diff too:)

Did as you wish, even though I think editing and reading the source files is easier when they are wrapped. And the rendered result is the same.

Tools/Platform_SDK/Building_Packages/README.md
```diff
@@ -0,0 +1,55 @@
...
+```
+
+Each build target lets you build software for a specific operating system version running on a specific hardware. Each target is listed under one tooling supporting the given OS version. Under each build target are listed existing snapshots of the target. So, in our example, we have installed one tooling (SailfishOS-4.4.0.58) and three targets for it (aarch64, armv7hl and i486). Each target has one snapshot (default).
+
+## Selecting build target
+
+The Platform SDK comes with a convenient tool for building packages, called `mb2`. It accepts target name with a command line option `--target` or `-t`. You can also set the default target using the `sb2-config` tool:
+```nosh
+$ sb2-config -d SailfishOS-4.4.0.58-aarch64
+```
```
martyone, line 32: I would try to keep sb2 behind the scene. Regular developers should not need to interact with it directly - we have mb2 build-shell [--maintain] and sdk-assistant maintain to execute random commands under the build environment without the need to remember the peculiar options of sb2.

````suggestion
The Platform SDK comes with a convenient tool for building packages, called `mb2`. It accepts build target name with the command line option `--target` (or briefly `-t`). You can also set the default target by creating a shell alias or by similar means:
```nosh
$ alias mb2='mb2 --target SailfishOS-4.4.0.58-aarch64'
```
````

File: Tools/Platform_SDK/Building_Packages/README.md [hunk 3]
```diff
@@ -0,0 +1,55 @@
+---
+title: Building Packages
+permalink: Tools/Platform_SDK/Building_Packages/
+parent: Sailfish Platform SDK
+layout: default
+nav_order: 300
+---
+
+This page deals specifically with building packages using the Platform SDK tools, i.e. `mb2`. If you are using the Sailfish SDK, please refer to the [Sailfish SDK documentation](/Tools/Sailfish_SDK/Building_packages) for information about how to use the `sfdk` tool for building packages.
```
sledges, line 9: > Did as you wish, even though I think editing and reading the source files is easier when they are wrapped. And the rendered result is the same.

I was of exact same opinion when I initially migrated the whole old wiki to be a hard-wrapped Markdown, and then @martyone pointed out that it's a nightmare to read the diff in PRs, to which I completely agreed.

Please see how easy it is to identify changes when text is not wrapped:
```suggestion
This page deals specifically with packages using the Platform. SDK tools, i.e. `mb2`. If you are using the Sailfish SDK, lorem ipsuum please refer to the [Sailfish SDK documentation](/Tools/Sailfish_SDK/Building_packages) duum for information. Rerum est ab suscipit ut non. About how to use the `sfdk` tool for building packages.
```
as opposed to the same change when hard-wrapped:
Screenshot from 2022-09-22 07-13-44

Also there will always be one contributor, who forgets to press M-x, causing maintainer to ask or re-wrap themselves before merging.

martyone, line 9: I would try to highlight here how closely the two environments are related and that the knowledge gained with one environment can be easily reused under the other.

````suggestion
If you have already learned how to build packages using command line interface of the regular [Sailfish SDK](/Tools/Sailfish_SDK/Building_packages), you will find the Platform SDK environment quite familiar. There is no surprise about because `sfdk`, the CLI frontend of the regular Sailfish SDK, is rather a lightweight dispatcher, redirecting your commands to the Platform SDK tools running behind the scene.

The most frequently used Platform SDK tool is `mb2`, and much of the `sfdk` commands [translate](https://github.com/sailfishos/sailfish-qtcreator/blob/master/share/qtcreator/sfdk/modules/20-building-mb2/manifest.json) directly to `mb2` invocations as in this example:
```nosh
(host) $ sfdk config --push OPTION1 VALUE
(host) $ sfdk config --push OPTION2
(host) $ sfdk SUBCOMMAND [ARGS...]
     ⇣
(PlatformSDK) $ mb2 --OPTION1 VALUE --OPTION2 SUBCOMMAND [ARGS...]
```

Here follows a brief introduction for those who jumps directly to Platform SDK use.
````

File: Tools/Platform_SDK/Building_Packages/README.md [hunk 4]
```diff
@@ -0,0 +1,55 @@
...
+cppqml-1.0-1.aarch64.rpm
+```
+
+The recipe for building the package can be found in a .spec file in the rpm directory, in our exacmple case `rpm/cppqml.spec`.
+
+## Further reading
+
+The `mb2` tool comes with comprehensive built-in documentation. It is accessible with command `mb2 --help`.
+
+
```
martyone, line 55: nitpick: trailing blank lines

Tools/Platform_SDK/Building_Packages/README.md
vigejolla: > Please see how easy it is to identify changes when text is not wrapped:

I agree that it would be nice if github could show these easier. But git (as in command line tool) handles it quite nicely, given some help: try `git diff --word-diff-regex=[^[:space:]]`

Tools/Platform_SDK/Building_Packages/README.md
vigejolla: I was wondering why setting the default target was made in such a weird way in all of our documentation. I mean, when there is a setting for it, it is funny not to use it and declare an alias instead. I would think those who use mb2 are quite capable developers, or at least they should be. They should be able to handle sb2-config as well. They shouldn't need training wheels.

vigejolla: That being said, why doesn't sdk-assistant have a command for setting the default target?

Tools/Platform_SDK/Building_Packages/README.md
vigejolla: removed

Tools/Platform_SDK/Building_Packages/README.md
martyone: `sb2-config` is limited to setting default target while the alias works for any of mb2's global options and somewhere else we encourage users to set some mb2 options from the start, so introducing `sb2-config` would be redundant.

`sdk-assistant` does not have a notion of default target because there are not much use cases for it using the default target. `sdk-assistant` is more often used to add/remove tools or work globally (like the register command). When someone needs to run a maintenance command under a particular target, it is often the target which is currently in use and it is more often desired to modify the snapshot and not the original, and `mb2 build-shell [--maintain]` takes care of all of this.

martyone (approved):

PR #120, opened 2022-09-23
Sideline direct sb2 usage

martyone:


Tools/Platform_SDK/README.md
```diff
@@ -10,56 +10,8 @@ nav_order: 200
...
-
-For the second goal: The platform SDK (not the host computer's OS) provides all the development tools and CPU-architecture-specific toolchains. It does not have any Sailfish OS library/header files (although it \*does\* have libraries/headers which can be used for building more tools).
-
-Finally, the third goal is met by having one or more "Target"s **inside** the platform SDK. A Target is a collection of libraries/headers for a specific version/CPU-architecture of Sailfish OS (or other platforms). This is the second level of nesting - the device environment - and access to it is via the [Scratchbox2](/Scratchbox2 "brokenlink") system (which is part of the SDK).
-
-How this affects things in practice:
-
-1.  The host machine does not need very much software installed into the native OS
-2.  Development tools are installed into the SDK container (and this is done the same way on Linux/Windows/Mac)
-3.  Build time dependencies for a given package (ie libraries and headers) are installed into the relevant Target container(s)
```
vigejolla, line 40: There's a lot of useful information here about the philosophy behind the Platform SDK. I don't think we should just drop it, but instead move it to a better place; I think the Toolchain page would be suitable for this kind of information. And we could also add similar information about the Sailfish SDK there.

vigejolla (approved):
LGTM. I added one suggestion there, but it's probably a little bit out of scope for this task and deserves a separate task.

Tools/Platform_SDK/README.md
martyone: Yup, it is mostly covered on other places already now:

https://github.com/sailfishos/docs.sailfishos.org/blob/jb58839/Tools/Platform_SDK/README.md#overview
https://github.com/sailfishos/docs.sailfishos.org/blob/jb58839/Tools/Sailfish_SDK/FAQ/README.md#why-are-you-using-virtual-machines
https://github.com/sailfishos/docs.sailfishos.org/blob/jb58839/Tools/Platform_SDK/Target_Installation/README.md#sailfish-platform-sdk-targets-and-toolings
https://github.com/sailfishos/docs.sailfishos.org/blob/jb58839/Reference/Toolchain/README.md#sailfish-sdk-toolchain

vigejolla: Yes, *mostly* covered, but still requires the reader to gather the bits and pieces from multiple places. It would be nice if we had one page which describes the whole picture on a high level, i.e. the philosophy of scratchbox2, the philosophy of platform SDK (the part which you just removed) and the philosophy of Sailfish SDK, and how they build on top of each other. It could make a nice bedtime story for someone ;)

Tools/Platform_SDK/Installation/README.md
```diff
@@ -8,7 +8,7 @@ nav_order: 100
 
 **Attention: Platform development is now more convenient with the regular [Sailfish SDK](/Develop/Platform) through its command line frontend 'sfdk', available on all Linux, macOS and Windows. The Linux-only chroot based Sailfish Platform SDK remains available for special applications. It also remains the default (documented) option for [Hardware Adaptation Development](/Tools/Hardware_Adaptation_Development_Kit), although the regular Sailfish SDK can be used for that purpose as well.**
 
-The Sailfish Platform SDK contains [Development Tools](/Tools/Development) like Scratchbox2, [MIC (Image Creation)](mic "brokenlink"), [Spectacle](/Spectacle "brokenlink"), osc, qemu, etc, to make it easier for a developer to work with Sailfish OS.
+The Sailfish Platform SDK contains [Development Tools](/Tools/Development) like mb2, [MIC (Image Creation)](mic "brokenlink"), osc, etc, to make it easier for a developer to work with Sailfish OS.
```
Thaodan, line 12: Why remove the mention of sb2? It is ok to favor using the SDK wrappers but hiding info doesn't help in my opinion.

Tools/Platform_SDK/README.md
Thaodan: I agree with @vigejolla however docs about what sb2 is can go to sb2 itself after a surface level briefing about what it does and is.

Tools/Sailfish_SDK/Building_packages/README.md
```diff
@@ -20,7 +20,7 @@ See the page about [deploying packages](/Tools/Sailfish_SDK/Deploying_packages)
 
 ## Building Packages Locally
 
-Once the user has installed the [Sailfish SDK](/Tools/Sailfish_SDK/Installation), they can build packages locally for the target architecture using the tool called `sfdk`. This tool provides a number of sub-commands wrapping around ScratchBox2 (`sb2`) and `rpmbuild` in a way which makes it simple to build packages from source repositories, given an RPM specification (.spec) file.
```
Thaodan, line 23: Doesn't that remove information? Sfdk wraps around more than just rpmbuild.

Tools/Platform_SDK/Installation/README.md
martyone: Scratchbox 2 has been invented with one main objective: to save the devs from the need to deal with the peculiarities of cross-compiling. We have gone one step further, saving the devs from the need to deal with the peculiarities of using sb2 :) . No one needs to be bothered with the `-m` and `-R` options of sb2 and what the various modes mean etc. in order to be able to build software for SFOS. When comes to executing arbitrary commands under the build environment, all that one needs to know is `mb2 build-shell [--maintain]` and `sdk-assistant maintain`. The original authors of Scratchbox 2 [stated it clearly](https://raw.githubusercontent.com/sailfishos/scratchbox2/master/docs/SB2_internals_1st_ed_20120425.pdf) that

> It should be noted that Scratchbox 2 is not an SDK (software development kit) itself. Instead, it is a cross-compilation toolkit; an engine which can be used to build SDKs.

and

> Most of the time Scratchbox 2 operates in the background. Some SDK implementations may decide
to hide it completely from the user.

Tools/Sailfish_SDK/Building_packages/README.md
martyone: This copes with commands doing something "[...] given an RPM specification (.spec) file.". But yeah I added the words "and others" there.

Tools/Platform_SDK/README.md
martyone: Yes, I will try to add the info somewhere.

martyone: Filed bug 58856 for that.

martyone:
Rebased

Tools/Platform_SDK/README.md
martyone: > Filed bug 58856 for that.

I mean I will merge this now or I am afraid there is a possibility to get in conflict with upcoming PRs easily.

PR #121, opened 2022-09-26
Add instructions for entering build target with mb2. JB#58824

vigejolla:


vigejolla:
> The build-shell vs maintenance mode concept seems interesting but I am afraid it doesn't fit the big image currently presented by the built-in docs. It might be that you missed the `--maintenance` option to `build-shell` and so got to the perception that there is such division.

Well, there *is* a division between the modes. But you are right, I somehow have missed the `build-shell --maintain` option, let's use that.

> I would also suggest to be a bit more specific about the intent and so saying "Running arbitrary build commands" instead of "Entering build environment".

Makes sense. I clearly wasn't able to let go of the original task description :roll_eyes: 

> Let me try to suggest a _skeleton_ based on copy-pasted content from the built-in help:

Thanks. There's good stuff there. Let's see what I can come up with that :)

Tools/Platform_SDK/Building_Packages/README.md
```diff
@@ -59,6 +59,51 @@ cppqml-1.0-1.aarch64.rpm
 
 The recipe for building the package can be found in a .spec file in the rpm directory, in our exacmple case `rpm/cppqml.spec`.
 
+## Running arbitrary commands
+
+Sometimes you might have the need to execute commands directly on the build target, instead of using the recipes specified in .spec files. You can do so in two modes: build-shell mode and maintentance mode.
```
martyone, line 64: "Do something on build target" is the common speech among us, but it is very inaccurate. It's not just the build target, but also the build tooling and the host that makes up the VFS presented to the application running under the sb2 session. The easiest approach to stay accurate is to limit use of the term "(build) target" and use the generic term "build environment" whenever possible.

martyone, line 64: I am afraid the build-shell mode and maintenance mode terms do not work well. It is the single build-shell command that allows you to run arbitrary commands under the build environment. And the build shell can be running in two modes - the default mode and the maintenance mode. So it would be OK to speak about opening the build-shell in maintenance mode, but I would prefer to not try to name the opposite (default) mode.

vigejolla: In this case I feel that the term "build environment" is too vague. Perhaps it could work if there was some additional description, but that would be beside the point here. Build target is more specific, and the fact that it is slightly inaccurate does not matter here.

vigejolla: I dropped the part talking about default mode.

martyone: It is not slightly inaccurate, it is very inaccurate. The command is not executed "on build target". It is executed under Scratchbox2 session.

> sb2  runs  command in a scratchbox2 session. If no command is given, an interactive shell is started.
>
> Scratchbox2 is typically used to create a  transparent  cross-compiling environment for developing programs and compiling software packages for e.g. embedded devices.
> [...]
>
> https://github.com/sailfishos/scratchbox2/blob/master/docs/sb2.1

Command run under Scratchbox2 session is presented with a VFS composed from parts of all the target, the tooling and the host file system trees. It is not just the target as if it was entered with e.g. chroot and we should not guide newcomers to adopt this wrong perception.

But we do not need to bother newcomers with the details, so we do not speak of sb2 sessions and how they work. We are very vague about that intentionally and so we use the commonly understood term "build environment" consistently to denote the environment where the build commands are executed (including commands that modify the build environment itself) as opposite to the host environment and the target environment (where the build results are deployed and run - the devices).

vigejolla: > But we do not need to bother newcomers with the details, so we do not speak of sb2 sessions and how they work. We are very vague about that intentionally and so we use the commonly understood term "build environment" consistently to denote the environment where the build commands are executed (including commands that modify the build environment itself) as opposite to the host environment and the target environment (where the build results are deployed and run - the devices).

I would accept that if that was true. But we speak of "build targets" in many places, exactly in situations like this. The term "build environment" is good when we compare the environment to the host environment. But now we want to make clear that we are not talking about the chroot environment, nor the tooling, both of which are also part of the build environment. Currently the text here is in line with e.g. `mb2 --help`.

martyone: I likely see what you want to express. That *this is (also) a way to modify what is stored under the build target tree*. But they are not asking how to modify what is under the build target tree. Their intent is to use/modify the build environment issuing commands directly. They are taught to use `build-shell [--maintain]` for that and that's all they need to know now. They do not need to know how/where the changes to the build environment are persisted (yet). They are also taught that there is a notion of a clean state of the build environment and that they can see the changes from the clean state with `build-requires diff` and revert their build environment to the clean state with `build-requires reset`. All this with no mention of the implementation details. Only in the detailed section of the built-in help they can learn that

> As explained later in the MANAGING SDK BUILD TOOLS section, any changes done under the *build environment* are persisted under *build targets*.

Now they learn that something is getting written to build targets. Only now and not earlier and only those who are interested. And those who adopt the high level usage patterns may never hit a need to learn.

The online docs is likely full of the legacy terminology, but the built-in docs has been cleaned up as much a possible in the sense that the low level details are only leaked when it is inevitable. We should do the same with new content on the online docs and ensure the consistency among the old content gradually. That's the only way forward.

vigejolla: Now, I see your point, and it seems you also see mine, so we can get forward :)

If they have read the pages in order so far, they are already familiar, or at least they know the existence of, with toolings and targets.

> As explained later in the MANAGING SDK BUILD TOOLS section, any changes done under the build environment are persisted under build targets.

I think I can use this so that we can both be happy about the result.

vigejolla: I changed the term to "build environment" and added a mention that the changes are persisted under build targets. I also added a link earlier in the page, where build targets are first mentioned, to the previous page which describes the toolings and targets.

martyone: yup, great! :)

Tools/Platform_SDK/Building_Packages/README.md
```diff
@@ -59,6 +59,47 @@ cppqml-1.0-1.aarch64.rpm
 
 The recipe for building the package can be found in a .spec file in the rpm directory, in our exacmple case `rpm/cppqml.spec`.
 
+## Running arbitrary commands
+
+Sometimes you might have the need to execute commands directly on the build environment. When you don't have a spec file, or you want to execute a command which is not in the spec, you can do so using the `mb2 build-shell` command. You can give the actual command you want to execute as an additional parameter.
```
martyone, line 64: It is not that unlikely scenario so I would suggest to reduce the conditional mood :)

Then, I may be wrong, but my feeling is saying "under" is more correct than "on" and "directly" should go as the last (build commands are always run directly under the build environment; our point here is to run the build commands under the build environment directly, not with the help of SPEC.)

```suggestion
When you don't have a spec file, or you want to execute a command which is not in the spec, you can use the `mb2 build-shell` command to enter the build environment and execute commands directly. You can give the actual command you want to execute as an additional parameter.
```

Tools/Platform_SDK/Building_Packages/README.md
```diff
@@ -59,6 +59,47 @@ cppqml-1.0-1.aarch64.rpm
...
+```
+
+If you leave the command out, an interactive shell is opened:
+```nosh
+$ mb2 build-shell
+NOTICE: Using the 'SailfishOS-4.4.0.58-aarch64.default' snapshot of the build target
+[SailfishOS-4.4.0.58-aarch64] $
+```
+
+Any changes done under the build environment are persisted under build targets. As you can see, by default the build-shell command uses the snapshots, just like the build commands.
```
martyone, line 80: At this moment we are not doing any changes to the build environment, so we should move this to the Maintenance mode session. I am afraid a newcomer will not be able to understand what "uses the snapshots" means (except users familiar with similar concepts like the `-snapshot` option to `qemu` :)), so we should say a bit more.

I would suggest to merge it to the paragraph starting "Unless the '--no-snapshot' global option..." and modify the paragraph a bit.

> Any changes done under the build environment are persisted under build targets. Unless the `--no-snapshot` option is used, mb2 uses a working copy (a "snapshot") of the actual build target to persist your build environment. You have seen the informal message regarding the use of a snapshot a few times already. Thanks to this you can revert any modifications to the build environment using the 'build-requires reset' command, so don't be afraid to experiment, but be aware that changes may get reset implicitly under certain conditions (read more in the built-in help of mb2).

vigejolla: I would rather link to the paragraph about snapshots, but unfortunately that PR is still in review, so I'll just accept your suggestion as an interim solution.

martyone (approved):

PR #123, opened 2022-09-27
Creating Help Articles folder

jpwalden:


Support/Help_Articles/README.md
```diff
@@ -0,0 +1,11 @@
+---
+title: Help Articles
+permalink: Support/Help_Articles
+has_children: true
+layout: default
+nav_order: 100
```
vigejolla, line 6: We now have same nav_order on Releases/README.md as well. It should be something different, I suggest 150 if we want to have this page between Releases and Supported Devices

vigejolla (changes requested):

vigejolla:
I don't fully see how this PR maps to the task 58506, are you sure the bug number in the branch name is correct?

On a related note, I don't think it makes sense to merge this PR without any "subpages". So, if your intention was indeed to add the page mentioned in task 58506 here, I suggest you add it in this same PR.

jpwalden:
> I don't fully see how this PR maps to the task 58506, are you sure the bug number in the branch name is correct?
> 
> On a related note, I don't think it makes sense to merge this PR without any "subpages". So, if your intention was indeed to add the page mentioned in task 58506 here, I suggest you add it in this same PR.

Yes, the intention was to add that ZD article, creating of the folder is the 1st step, therefore we used that ID.

vigejolla:
Please squash the commits

vigejolla (approved):
LGTM, but please don't merge this before you have some content for the folder as well

jovirkku:
The created folder Support/Help_Articles will contain plenty of subfolders, each for one help article.

jovirkku (changes requested):
Line 142 in README.md: the code has "Sailfish-dir-content-3.1.0.11.PNG" while the picture has the extension .png ==> the picture is not shown in chapter 4 (picture 4b).

Line 104 has "...with Android O****S."

Chapter 4: "...is always available in [this repository](https://shop.jolla.com/downloads/) [*]"
As that sentence has the hyperlink then perhaps the "[*]" and the later (two lines down) URL are not necessary.

The header numbers, then. I think this DOC service has an automatic TOC and numbering of headers. The manually written numbers might lead to duplicate numbering. Not sure, though. 

Otherwise, it looks good.

Support/Help_Articles/Installing_Sailfish/README.md
```diff
@@ -0,0 +1,190 @@
+---
+title: Reinstalling Sailfish OS
+permalink: Support/Help_Articles/Installing_Sailfish
+parent: Help_Articles
```
vigejolla, line 4:
```suggestion
parent: Help Articles
```

parent should have the title of the parent page, not directory name

Support/Help_Articles/Installing_Sailfish/README.md
```diff
@@ -0,0 +1,190 @@
+---
+title: Reinstalling Sailfish OS
```
vigejolla, line 2: I would prefer having title identical to the directory name (not counting the ' ' vs '_' difference). So I would suggest you change either the title to "Installing Sailfish" or change the directory name to "Reinstalling_Sailfish_OS". In my opinion the latter is the better option.

vigejolla (changes requested):
Please fix also those findings from @jovirkku

Support/Help_Articles/Installing_Sailfish/README.md
jpwalden: Done

Support/Help_Articles/Installing_Sailfish/README.md
jpwalden: Directory name changed to Reinstalling_Sailfish_OS

vigejolla (approved):
LGTM

PR #126, opened 2022-10-04
Tips and tricks

jovirkku:
Several ZD articles combined into Tips and Tricks. JB#58163.


jpwalden (commented):
These sound little bit odd, but very minor (line 73):
"Should you receive an error about an incorrect password or username.."
Could be: "If you receive an error.."
same with (line 333) "Should you have problems with installing.."
Could be "If you have problems.."

Some of the links are not functional:
Lines: 246, 320, 371, 377

Otherwise looks good to me.

Support/Help_Articles/Tips_and_Tricks/README.md
```diff
@@ -0,0 +1,406 @@
...
+9.  Should the **charge level of the battery** go down faster than expected then read **[this article](https://jolla.zendesk.com/hc/en-us/articles/201922397)** and consider the proposed different ways of saving energy.
+
+
+# Forced power off
+
+Your phone might end up in a situation where it is stuck in such a way that turning it off in the usual way with the Power key is not possible.
+
+The device manufacturer has created the following backup method for forcing the phone to the off state:
+
+1.  Press both the Power and the Volume Up keys down at the same time and keep them pressed.
```
spiiroin, line 223: Does this like this work for all currently supported devices? (e.g. in n9 or n900 it was enough just to keep power key pressed for 7 seconds or stte, but maybe some sort of "industry standard" has risen since those days?)

jovirkku: It appears to work on all Xperia phones we support.

spiiroin: To me the text just sounds a bit too definite since there is some variance & we do provide device specific instructions elsewhere. On a quick search android boards seem to use language to the effect of: "most devices power+vol down; some power + both vol keys; rare instances something else". Does not matter if it works as described all devices we expect people to use, but might be worthwhile to name few models where we know this info to be true? (but I'll leave it up to you to decide whether it is worthwhile, it is just fine as it is)

jovirkku: I have successfully used that for years in my Care tasks. Many customers have been helped with that advice.
I tried it now on Xperia 10 III, Xperia 10 II, Xperia XA2, and Xperia X. I was not able to force those phones to an abnormal state but, instead, pressing the keys at the security code prompt led to those 3 plays of the vibrator, after which the phones are off. At least Xperia 10 III and II do not play the triple vibra sequence if Power + VolDown are pressed.
I trust on my sentence.

spiiroin (approved):
My concerns were addressed.

vigejolla:
please rebase on top of master

Support/Help_Articles/Tips_and_Tricks/README.md
```diff
@@ -0,0 +1,406 @@
+---
+title:
+permalink: Support/Help_Articles/Tips_and_Tricks
+has_children: true
+layout: default
+nav_order: 1000
```
vigejolla, line 6: missing parent

Support/Help_Articles/Tips_and_Tricks/README.md
```diff
@@ -0,0 +1,406 @@
+---
+title:
+permalink: Support/Help_Articles/Tips_and_Tricks
+has_children: true
```
vigejolla, line 4: no, this document does not have children

Support/Help_Articles/Tips_and_Tricks/README.md
```diff
@@ -0,0 +1,406 @@
...
+
+* Go to the menu on Settings > Accounts
+* Tap "Add account"
+
+|<a href="New-account.png" style="width:30em;display:block">
+<img src="New-account.png"
+alt="Adding an account"
+class="md_thumbnail" style="max-width:100%"/>
+</a>|
+|-|
```
vigejolla, line 41: Not sure what you are trying to accomplish here but it clearly isn't working

vigejolla (changes requested):
Some issues need fixing before this can be merged

Support/Help_Articles/Tips_and_Tricks/README.md
```diff
@@ -0,0 +1,406 @@
+---
+title:
+permalink: Support/Help_Articles/Tips_and_Tricks
```
vigejolla, line 3:
```suggestion
permalink: Support/Help_Articles/Tips_and_Tricks/
```

jovirkku: Slash added.

Support/Help_Articles/Tips_and_Tricks/README.md
jovirkku: Inserting pictures into 3x1 tables fixed. The main declaration must be on one row (from <a href to </a>).

jovirkku:
Done so far:
- title added
- parent added
- children removed
- slash added to the end of permalink
(I list the changes here as I cannot find how to handle the "unresolved conversations", some of which have "outdated")

PR #136, opened 2022-10-12
Testing advice page

jpwalden:

### DIFF
--- a/Develop/Platform/Testing_Advice/README.md
+++ b/Develop/Platform/Testing_Advice/README.md
@@ -1,0 +1,127 @@
+---
+title: Testing Advice
+permalink: Develop/Platform/Testing_Advice/
+has_children: false
+parent: Platform
+layout: default
+nav_order: 200
+---
+
+# Testing in general
+
+Main point of testing is to find bugs, in case bug is found the reporting of the bug is crucial. It is essential that bug report includes all the needed information, there fore bug report template has to be used and all the points needs to be filled:
+```
+REPRODUCIBILITY (% or how often the bug can be seen):
+BUILD ID:
+BUILD FLAVOR (devel/release):
+HARDWARE (Jolla C, Xperia XA2, Xperia X10 II, ...):
+DID IT WORK CORRECTLY IN THE PREVIOUS PUBLIC RELEASE:
+DESCRIPTION:
+PRECONDITIONS:
+STEPS TO REPRODUCE:
+1)
+2) 
+3) 
+EXPECTED RESULT:
+ACTUAL RESULT:
+ADDITIONAL INFORMATION: (Please ALWAYS attach relevant data such as logs, screenshots, etc...)
+SUMMARY OF SOLUTION: (After resolving this bug/task, please write a short summary on the design decisions taken, eventually, and how the solution helps to improve Sailfish OS in bigger picture. If you see it better to write the summary to a comment, please do so)
+```
+
+When evaluating the found bug it is important to make sure that the bug is valid, good way is to test if the bug is visible with another device (if available), also reference device testing gives valuable infromation. Naturally this depends on the bug type, since some bugs are random and can't be easily reproduced, some are clearly visible bugs which can be seen simply from the UI and so on.
+
+In case bug can be reproduced easily then logs are not that crucial, since then the developer can easily also get those if the steps to reproduce are clear or sometimes logs are not simply needed if the bug is related to UI glitch etc. In case bug is hard to reproduce then it is very important to provide also logs since it might be that the bug simply can't be reproduced due to test environment etc. reasons. When providing logs it is also important either study the logs and highlight the related lines or at least inform on what time the bug orccurs so that it is easier to find the correct lines from the logs.
+
+When providing logs at least [journal logs](https://jolla.zendesk.com/hc/en-us/articles/202886373) should be provided.
+
+On some areas journal logs are not enough to reveal the root cause and therefore additional logs might be needed.
+More information about logs below:
+
+# Collecting debug information
+[This content is copied from forum](https://forum.sailfishos.org/t/wiki-collecting-debug-information/12751)
+
+## Basics
+
+* Enable developer mode
+* Show failed services: `systemctl list-units --failed`
+* Collect logs for a specific systemd unit `journalctl -u <unitname>`
+* Follow actions of out of memory killer: `dmesg -w | egrep "lowmemorykiller|oom_reaper|^ {19}.*|Out of memory"`
+* `collect-logs.zip` from [Collecting basic logs from a Sailfish device](https://jolla.zendesk.com/hc/en-us/articles/360013910599-Collecting-basic-logs-from-a-Sailfish-device) 
+
+## Audio
+* `systemctl status pulseaudio` and `journalctl -u pulseaudio`
+
+## Bluetooth
+* `systemctl status mpris` mpris is responsible for actions (play,pause,…)
+* `systemctl status bluetooth` and `journalctl -u bluetooth`
+
+## Android
+* `systemctl status aliendalvik` and `journalctl -u aliendalvik`
+
+## Android apps
+* `/system/bin/logcat`
+* `devel-su lxc-attach -n aliendalvik /system/bin/logcat`
+
+## EMail
+* [How to collect EMail logs](https://jolla.zendesk.com/hc/en-us/articles/201975906-How-to-collect-Email-logs-General-email-IMAP-POP-)
+
+## Mobile Data / Calling / SMS
+the logs may contain your personal data!
+* logs `journalctl -u ofono`
+* sms `devel-su journalctl --system -f | grep jolla-messages`
+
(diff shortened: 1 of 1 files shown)


Develop/Platform/testing_advice/README.md
```diff
@@ -0,0 +1,90 @@
+---
+title: Testing Advice
+permalink: Develop/Platform/testing_advice/
```
vigejolla, line 3: Please use Capital first letter in directory name

Develop/Platform/testing_advice/README.md
```diff
@@ -0,0 +1,90 @@
...
+has_children: true
+parent: Platform
+layout: default
+nav_order: 150
```
vigejolla, line 7: It has been our custom to use full hundreds unless adding a page between two already existing pages

Develop/Platform/testing_advice/README.md
```diff
@@ -0,0 +1,90 @@
+---
+title: Testing Advice
+permalink: Develop/Platform/testing_advice/
+has_children: true
```
vigejolla, line 4: Should be false as this page does not have children

Develop/Platform/testing_advice/README.md
```diff
@@ -0,0 +1,90 @@
...
+parent: Platform
+layout: default
+nav_order: 150
+---
+
+#Testing in general
```
vigejolla, line 10: Missing space between # and T

vigejolla: same for all headings

Develop/Platform/testing_advice/README.md
```diff
@@ -0,0 +1,90 @@
...
+
+Main point of testing is to find bugs, in case bug is found the reporting of the bug is crucial. It is essential that bug report includes all the needed information, there fore bug report template has to be used and all the points needs to be filled:
+```nosh
```
vigejolla, line 13: No need for `nosh` here. It's only used for shell script snippets in order to disable code highlighting for those. The reason for that is that the code highlighting doesn't quite do what we would like it to do on shell scripts.

Develop/Platform/testing_advice/README.md
```diff
@@ -0,0 +1,90 @@
...
+1)
+2) 
+3) 
+EXPECTED RESULT:
+ACTUAL RESULT:
+ADDITIONAL INFORMATION: (Please ALWAYS attach relevant data such as logs, screenshots, etc...)
+SUMMARY OF SOLUTION: (After resolving this bug/task, please write a short summary on the design decisions taken, eventually, and how the solution helps to improve Sailfish OS in bigger picture. If you see it better to write the summary to a comment, please do so)
+```
+
+When evaluating the found bug it is important to make sure that the bug is valid, good way is to test if the bug is visible with another device (if available), also reference device testing gives valuable infromation. Naturally this depeneds on the bug type, since some bugs are random and can't be easily reproduced, some are clearly visible bugs which can be seen simply from the UI and so on.
```
jovirkku, line 31: depeneds > depends

Develop/Platform/testing_advice/README.md
```diff
@@ -0,0 +1,90 @@
...
+
+In case bug can be reproduced easily then logs are not that crucial, since then the developer can easily also get those if the steps to reproduce are clear or sometimes logs are not simply needed if the bug is related to UI glitch etc. In case bug is hard to reproduce then it is very important to provide also logs since it might be that the bug simply can't be reproduced due to test environment etc. reasons. When providing logs it is also important either study the logs and highlight the related lines or at least inform on what time the bug orccurs so that it is easier to find the correct lines from the logs. 
+
+When providing logs at least journal logs (https://jolla.zendesk.com/hc/en-us/articles/202886373) should be provided
+
+On some areas journal logs are not enough to reveal the root cause and there fore additional logs might be needed.
```
jovirkku, line 37: there fore > therefore

jovirkku (changes requested):

Develop/Platform/testing_advice/README.md
```diff
@@ -0,0 +1,90 @@
...
+
+
+#Types of testing conducted, Schedule and milestones, Criteria for starting and ending of testing.
+
+Type of the test cases is defined case-by-case with the developer, but usually we are using use case –type of test cases if possible. If the feature can’t be tested from the UI then the testing method is defined with the developer.
+
+Schedule and milestones are also defined case-by-case since some features need much more testing than others and some features need specific testing environment which needs to be build by the tester before the testing can be started.
+
+Testing is started as soon as the basic functionality of the feature is in place. Preferably before the code is merged to devel and testing will be continued when the code is in devel.
```
jovirkku, line 46: I wonder if this is clear for an external reader. Especially, what is 'devel'. Should there be a short description of our development lifecycle, or at least explain that 'devel' means the development codebase (or branch?)  to which developers merge their work? Secondly, I wonder how a feature can be tested before it is in 'devel'. Would the tester work in the developer's environment (or in the copy of that)?

Develop/Platform/testing_advice/README.md
```diff
@@ -0,0 +1,90 @@
...
+```
+
+When evaluating the found bug it is important to make sure that the bug is valid, good way is to test if the bug is visible with another device (if available), also reference device testing gives valuable infromation. Naturally this depeneds on the bug type, since some bugs are random and can't be easily reproduced, some are clearly visible bugs which can be seen simply from the UI and so on.
+
+In case bug can be reproduced easily then logs are not that crucial, since then the developer can easily also get those if the steps to reproduce are clear or sometimes logs are not simply needed if the bug is related to UI glitch etc. In case bug is hard to reproduce then it is very important to provide also logs since it might be that the bug simply can't be reproduced due to test environment etc. reasons. When providing logs it is also important either study the logs and highlight the related lines or at least inform on what time the bug orccurs so that it is easier to find the correct lines from the logs. 
+
+When providing logs at least journal logs (https://jolla.zendesk.com/hc/en-us/articles/202886373) should be provided
```
jovirkku, line 35: "It's good practice to surround URLs: <http://some.url>"
(quoting from https://github.com/sailfishos/docs.sailfishos.org/#sailfish-os-documentation)

Develop/Platform/testing_advice/README.md
jpwalden: Corrected

Develop/Platform/testing_advice/README.md
jpwalden: Changed to 200

Develop/Platform/testing_advice/README.md
jpwalden: Corrected

Develop/Platform/testing_advice/README.md
jpwalden: Corrected

Develop/Platform/testing_advice/README.md
jpwalden: Removed

Develop/Platform/testing_advice/README.md
jpwalden: Corrected

Develop/Platform/testing_advice/README.md
jpwalden: Corrected

Develop/Platform/testing_advice/README.md
jpwalden: Didn't change this at least yet. Added more content on the log part mainly from the https://forum.sailfishos.org/t/wiki-collecting-debug-information/12751

Develop/Platform/Testing_Advice/README.md
```diff
@@ -0,0 +1,127 @@
...
+
+In case bug can be reproduced easily then logs are not that crucial, since then the developer can easily also get those if the steps to reproduce are clear or sometimes logs are not simply needed if the bug is related to UI glitch etc. In case bug is hard to reproduce then it is very important to provide also logs since it might be that the bug simply can't be reproduced due to test environment etc. reasons. When providing logs it is also important either study the logs and highlight the related lines or at least inform on what time the bug orccurs so that it is easier to find the correct lines from the logs.
+
+When providing logs at least [journal logs](https://jolla.zendesk.com/hc/en-us/articles/202886373) should be provided.
+
+On some areas journal logs are not enough to reveal the root cause and therefore additional logs might be needed.
+More information about logs below:
+
+# Collecting debug information
+[This content is copied from forum](https://forum.sailfishos.org/t/wiki-collecting-debug-information/12751)
```
vigejolla, line 41: I think we can leave this bit out from the page, as the reader is probably not that interested where the information is from. I would instead mention that in the commit message.

vigejolla:
To me this looks good, but I'll give @jovirkku a chance to verify himself that his concerns have been addressed. Please rebase on top of master.

jovirkku (approved):
Looks ok to me.

vigejolla (approved):

PR #146, opened 2022-10-26
Add dynamic table of contents to Support and Help Articles pages

vigejolla:


jovirkku:
What is the problem we are trying to solve in this PR? Aren't the TOCs dynamic/automatic already?

vigejolla:
> What is the problem we are trying to solve in this PR? Aren't the TOCs dynamic/automatic already?

Try going to https://docs.sailfishos.org/Support/ on your mobile device in portrait mode and you will surely see

vigejolla:
> What is the problem we are trying to solve in this PR? Aren't the TOCs dynamic/automatic already?

Support page before changes:
support-before

Support page after changes: 
support-after

similarly, help page before changes:
help-before
and after changes:
help-after

jovirkku:
Yes, understood. That is not user-friendly.  

Although, if one happens to tap the hamburger button (top right) in the situation of the first picture (i.e., https://docs.sailfishos.org/Support/) then the Help Articles item with the expansion arrow appears. The documents list can be seen.

But yes, I am in favour of a better solution.

rainemak (approved):
As @jovirkku verbally approved already I'll approve formally as well.

jpwalden:
Thanks, looks good to me as well, I had the same question that Jorma had, but that was very well answered -thanks for that.

PR #152, opened 2022-11-02
Make the first paragraph a bit calmer

vigejolla:
The bold "attention" text was confusing new HW adaptiation developers. I also dropped the mention of using the Sailfish SDK for HW adaptation. While it's technically possible to do so, we haven't documented that. Until we have it documented, it probably doesn't make sense to advertise it.

### DIFF
--- a/Tools/Platform_SDK/Installation/README.md
+++ b/Tools/Platform_SDK/Installation/README.md
@@ -6,9 +6,9 @@ layout: default
 nav_order: 100
 ---
 
-**Attention: Platform development is now more convenient with the regular [Sailfish SDK](/Develop/Platform) through its command line frontend 'sfdk', available on all Linux, macOS and Windows. The Linux-only chroot based Sailfish Platform SDK remains available for special applications. It also remains the default (documented) option for [Hardware Adaptation Development](/Tools/Hardware_Adaptation_Development_Kit), although the regular Sailfish SDK can be used for that purpose as well.**
+The Sailfish Platform SDK is the default (documented) option for [Hardware Adaptation Development](/Tools/Hardware_Adaptation_Development_Kit). Further, it can be used for platform development.
 
-The Sailfish Platform SDK contains [Development Tools](/Tools/Development) like mb2, [MIC (Image Creation)](https://github.com/sailfishos/mic/), [osc](https://github.com/sailfishos-mirror/osc/), etc, to make it easier for a developer to work with Sailfish OS.
+Note: Platform development is now more convenient with the regular [Sailfish SDK](/Develop/Platform) through its command line frontend 'sfdk', available on all Linux, macOS and Windows.
 
 # Quick start
 
--- a/Tools/Platform_SDK/README.md
+++ b/Tools/Platform_SDK/README.md
@@ -6,7 +6,9 @@ layout: default
 nav_order: 200
 ---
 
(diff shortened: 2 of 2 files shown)


martyone (approved):
There is one more occurrence of this paragraph in /Tools/Platform_SDK/Installation/. Otherwise LGTM.

Tools/Platform_SDK/README.md
```diff
@@ -6,7 +6,9 @@ layout: default
 nav_order: 200
 ---
 
-**Attention: Platform development is now more convenient with the regular [Sailfish SDK](/Develop/Platform) through its command line frontend 'sfdk', available on all Linux, macOS and Windows. The Linux-only chroot based Sailfish Platform SDK remains available for special applications. It also remains the default (documented) option for [Hardware Adaptation Development](/Tools/Hardware_Adaptation_Development_Kit), although the regular Sailfish SDK can be used for that purpose as well.**
+Note: Platform development is now more convenient with the regular [Sailfish SDK](/Develop/Platform) through its command line frontend 'sfdk', available on all Linux, macOS and Windows.
```
rainemak, line 9: Better but I can still see how this leads to wrong paths. This is the first thing that you see here. I could imagine that new hw adaptation dev easily starts pondering here. Install Sailfish SDK and continues with HA development and soon all goes south and a lot of questions bubbles up.

Maybe just swapping the Note and next paragraph would be enough.

Tools/Platform_SDK/Installation/README.md
```diff
@@ -6,10 +6,12 @@ layout: default
 nav_order: 100
 ---
 
-**Attention: Platform development is now more convenient with the regular [Sailfish SDK](/Develop/Platform) through its command line frontend 'sfdk', available on all Linux, macOS and Windows. The Linux-only chroot based Sailfish Platform SDK remains available for special applications. It also remains the default (documented) option for [Hardware Adaptation Development](/Tools/Hardware_Adaptation_Development_Kit), although the regular Sailfish SDK can be used for that purpose as well.**
+The Sailfish Platform SDK is available for special applications. It is also the default (documented) option for [Hardware Adaptation Development](/Tools/Hardware_Adaptation_Development_Kit).
```
rainemak, line 9: What is that "special applications"? What if this would be just "The Sailfish Platform SDK is available for platform development."

martyone: That would imply Sailfish Platform SDK is needed for platform development, which is not true. "Special applications" has the usual meaning - if you don't know why to use it, you don't need it. If you are told to use it, you don't need to ask.

rainemak: No, for newbie "special applications" does not have usual meaning. The "special application" just causes confusion.

How about
```
The Sailfish Platform SDK is the default (documented) option for [Hardware Adaptation Development](/Tools/Hardware_Adaptation_Development_Kit). Further, it can be used for platform development.
```
After this the Note can steer devs towards Sailfish SDK if they so see.

martyone: Works for me but please drop the "The Sailfish Platform SDK contains [Development Tools](/Tools/Development) like mb2 ..." paragraph then, which distracts the flow and has little sense here on the installation page.

martyone: The other file needs to be updated too.

vigejolla: updated the other file as well

rainemak (approved):

PR #158, opened 2022-11-07
Creating of Internet sharing article

jpwalden:


jovirkku (commented):
This document looks good to me.

Before making it, we (JP and I) discussed if we keep the "USB tethering" section of it or not. It was unclear if that still works or not. It seems to work, I noticed. However, making it work requires some effort on the command line. As this feature is not advertised on the Sailfish UI, perhaps we can indeed drop it from this document at this point. 

- Any other thoughts with some reasoning? @vigejolla

vigejolla:
> This document looks good to me.
> 
> Before making it, we (JP and I) discussed if we keep the "USB tethering" section of it or not. It was unclear if that still works or not. It seems to work, I noticed. However, making it work requires some effort on the command line. As this feature is not advertised on the Sailfish UI, perhaps we can indeed drop it from this document at this point.
> 
>     * Any other thoughts with some reasoning? @vigejolla

I've seen a few people trying to use it, clearly it's something people want. I would keep it, but perhaps start with something along the lines of "Sharing over WLAN is the preferred method, but if you absolutely have to use USB tethering, this is how you can do it".

jpwalden:
USB tethering part added

jovirkku (approved):
This document looks ok to me now, after adding the USB tethering part. There is the recommendation to primarily use WLAN tethering.

Support/Help_Articles/Internet_Sharing_JB59215/README.md
```diff
@@ -0,0 +1,97 @@
+---
+title: Internet Sharing
+permalink: Support/Help_Articles/Internet_Sharing/
+parent: Help Articles
+layout: default
+nav_order: 85
```
vigejolla, line 6: Two issues:
 - Don't put the JB number to the path name
 - Internet Sharing should be before "Making Backup", which has nav_order 80

vigejolla (approved):

PR #157, opened 2022-11-07
Update web-link to stale content at TJC

Olf0:
While I feel honored by the fact that @jpwalden mentions my "Guide for installing Sailfish X on Sony Xperias" in his [PR 141 "Creating Flashing Troubleshooting article"](https://github.com/sailfishos/docs.sailfishos.org/pull/141/files#diff-148bb7402f0f46f63163e1c3a572b9bebbd811e9d65dcf58e175b79e03d27d24R95), it links to [an outdated version of this guide (the one at TJC)](https://together.jolla.com/question/222126/), because TJC was set to "read only" at the end of 2020, hence the version there cannot be updated ever since.  
Thus the canonical URL for this guide is [https://gitlab.com/Olf0/sailfishX#guide-installing-sailfishx-on-xperias](https://gitlab.com/Olf0/sailfishX#guide-installing-sailfishx-on-xperias), where it is still maintained.

This PR solely swaps the stale URL with the current one.

PR #161, opened 2022-11-07
Shorten & rephrase two sentences into one

Olf0:
Originally I aimed at rectifying the "*(read-only)*" leftover from the former web-link to TJC (and missed in PR #157), plus the awkward phrase "beyond a link": At GitHub "Issues" and PRs are open, hence "read-only" does not fit.

Ultimately I ended up joining and shortening these two sentences, while trying to retain the style by linking only single words in bold.

Additionally I slightly altered the start of a later sentence to denote that different sections require varying skills.


vigejolla (approved):
LGTM, thanks again

PR #166, opened 2022-11-09
Creating of Android app support article

jpwalden:


attah:
Under limitations, maybe also mention NFC, WiFi settings/scanning, VPN config and USB device access.

Support/Help_Articles/Android_Application_Support/README.md
```diff
@@ -0,0 +1,181 @@
...
+title: Android Application Support
+permalink: Support/Help_Articles/Android_Application_Support/
+parent: Help Articles
+layout: default
+nav_order: 40
+---
+
+Sailfish OS aims to be compatible with the majority of Android apps, but due to the sheer volume of Android applications designed for Android-based devices with different configurations, we cannot guarantee 100% compatibility.
+
+In this article, you will learn how to make the best of the **AndroidTM App Support** on your Sailfish device.
```
jovirkku, line 11: I wonder if the trademark could be written as follows: Android<sup>TM</sup> App Support.
Is there a better way for GitHub markdown?

vigejolla: Nothing wrong with `<sup>TM</sup>`, but another option is to just use the ™ character. Feel free to copy/paste it from here if your keyboard doesn't happen to have it. The result looks slightly different, I don't know which one is better.

jovirkku:
> Under limitations, maybe also mention NFC, WiFi settings/scanning, VPN config and USB device access.

The problems around AAS are probably not limited to these. One-quarter of our Bugzilla is about AAS. I would recommend that we keep this document as a generic one, on the level of Android generations, and even remove the current 2 examples (BT and SMS/calendar).  Or else this document would grow big.

Support/Help_Articles/Android_Application_Support/README.md
```diff
@@ -0,0 +1,181 @@
+---
+title: Android Application Support
+permalink: Support/Help_Articles/Android_Application_Support/
+parent: Help Articles
+layout: default
+nav_order: 40
```
jovirkku, line 6: nav_order should be 10 to force this doc to the 2nd place from the top (Android App Settings is going on top with 7).

vigejolla: Perhaps we should plase Android App Settings under Android Application Support? I mean, like `Support/Help_Articles/Android_Application_Support/Android_App_Settings/` ? Just and idea...

jpwalden: ok, thanks. Need to fix that, just pushed the other changes here, but didn't see this one in time.

jpwalden: nav order changed to 10

Support/Help_Articles/Android_Application_Support/README.md
jpwalden: Thanks for the recommendations, used the ™ character.

jpwalden:
> > Under limitations, maybe also mention NFC, WiFi settings/scanning, VPN config and USB device access.
> 
> The problems around AAS are probably not limited to these. One-quarter of our Bugzilla is about AAS. I would recommend that we keep this document as a generic one, on the level of Android generations, and even remove the current 2 examples (BT and SMS/calendar) - lines 76-77 and the last two words of line 74. Or else this document would grow big.

Thanks for the comments! For now decided to follow this guideline

Support/Help_Articles/Android_Application_Support/README.md
jovirkku: > Perhaps we should place Android App Settings under Android Application Support? I mean, like Support/Help_Articles/Android_Application_Support/Android_App_Settings/ ? Just and idea..

That might work. There is at least one more "Android" doc coming up.

jovirkku:
LGTM now.

vigejolla (approved):

Thaodan:
jovirkku ***@***.***> writes:

> I wonder if the trademark could be written as follows: Android<sup>TM</sup> App Support.
> Is there a better way for GitHub markdown?

We can use the UTF-8 symbol like this Android™.

Thaodan:
Ville Nummela ***@***.***> writes:

> +In this article, you will learn how to make the best of the **AndroidTM App Support** on your Sailfish device.
>
> Nothing wrong with `<sup>TM</sup>`, but another option is to just use
> the ™ character. Feel free to copy/paste it from here if your keyboard
> doesn't happen to have it. The result looks slightly different, I
> don't know which one is better.

Ah to late :)

PR #169, opened 2022-11-10
Make the title of Backup-Restore match with permalink

jovirkku:
I want to modify the title of the document so that it would be similar to the permalink (folder path). Different names cause confusion.


jovirkku:
This is JB#59278.

vigejolla (approved):

PR #172, opened 2022-11-11
Creating Backup extraction document

jpwalden:


jpwalden:
Just realized that the nav order should be 14 not 12, but before fixing that lets see if there is something else to be fixed.

vigejolla:
> Just realized that the nav order should be 14 not 12, but before fixing that lets see if there is something else to be fixed.

I don't see anything else wrong

jpwalden:
> > Just realized that the nav order should be 14 not 12, but before fixing that lets see if there is something else to be fixed.
> 
> I don't see anything else wrong

Ok -thanks! fixed that.

vigejolla (approved):

PR #173, opened 2022-11-11
Unify references to RPM SPEC files

vigejolla:


martyone (approved):
I would consider to mention "RPM" with "SPEC" wherever it occurs first in a paragraph/section where the context has not been introduced otherwise.

vigejolla:
> I would consider to mention "RPM" with "SPEC" wherever it occurs first in a paragraph/section where the context has not been introduced otherwise.

That's a good idea, I'll do that.

PR #186, opened 2022-11-23
Bluetooth pairing

jovirkku:


jovirkku:
I have no more changes for this in my mind.

jpwalden (approved):
Document looks good to me, however we have now ran out of the numbers on the nav order, the number 14 is already in use and numbers around it as well, so those needs to be adjusted

vigejolla:
> Document looks good to me, however we have now ran out of the numbers on the nav order, the number 14 is already in use and numbers around it as well, so those needs to be adjusted

It's even worse, we have accidentally already used 14 twice, for both "Backup Extraction" and "Battery Lifetime". @jovirkku could you make some room here? Preferably in a separate commit.

jovirkku:
Android App Support: 10
Backup Extraction: 14 (NOTE)
Backup and Restore: 13  (NOTE)
Battery Lifetime: 14
Bluetooth Pairing: 14
Collecting logs with Logcat: 15

It has been "broken" already before my commit. Fixing the indexes requires touching 10-20 documents. But it is doable, of course.

vigejolla:
> It has been "broken" already before my commit. Fixing the indexes requires touching 10-20 documents. But it is doable, of course.

That is well understood. Still, it would be easiest to just fix it here instead of opening a new task for it. But for the sake of clarity it's still best to make those changes in a separate commit.

jovirkku:
> > It has been "broken" already before my commit. Fixing the indexes requires touching 10-20 documents. But it is doable, of course.
> 
> That is well understood. Still, it would be easiest to just fix it here instead of opening a new task for it. But for the sake of clarity it's still best to make those changes in a separate commit.

https://bz.jollamobile.com/show_bug.cgi?id=59444

vigejolla:
> https://bz.jollamobile.com/show_bug.cgi?id=59444

Ok, let's finish that one first and then rebase this PR on top of that, if that's ok with you?

jovirkku:
> > https://bz.jollamobile.com/show_bug.cgi?id=59444
> 
> Ok, let's finish that one first and then rebase this PR on top of that, if that's ok with you?

Yes, ok.

jovirkku:
Nav_order fixed.

vigejolla (approved):

PR #192, opened 2022-11-28
Move files between Mac and Sailfish

jovirkku:


jpwalden (approved):
Looks good to me, just started to think that if we want to keep the "thanks" -texts for the community contributors should we list those in the end of the document, so that those are not in the "middle" of the document?

vigejolla:
> Looks good to me, just started to think that if we want to keep the "thanks" -texts for the community contributors should we list those in the end of the document, so that those are not in the "middle" of the document?

I agree. I think the best would probably be adding them as footnotes.

jovirkku:
Thanks to the community guys are now shown as footnotes.

vigejolla (approved):

PR #197, opened 2022-11-30
Adding remove AAS. Rearranging Android App Support folders.

jovirkku:


jpwalden (approved):
looks good to me

Support/Help_Articles/Android_App_Support/README.md
```diff
@@ -0,0 +1,22 @@
+---
```
vigejolla, line 1: I think this "Android App Support contents page" is not necessary. We already have "Android Application Support" page, we could just use that and add the other pages under it. That's what we have done elsewhere. We have used these "contents" pages only in those cases where we don't have actual content for the top level document.

Some examples of having subpages under a page: https://docs.sailfishos.org/Reference/I18n/ https://docs.sailfishos.org/Reference/Core_Areas_and_APIs/ https://docs.sailfishos.org/Tools/Platform_SDK/

Having only list of subpages as the content of a page is the exception and not the rule

Support/Help_Articles/Android_App_Support/Removing_Android_App_Support/README.md
```diff
@@ -0,0 +1,65 @@
...
+cd /home/defaultuser          ## On some phones:  cd /home/nemo
+rm -rf /home/.android/*
+rm -rf android_storage
+reboot
+```
+3) Restart the device.
+
+# Factory reset or reinstalling Sailfish OS
+
+**[Factory reset](https:///Support/Help_Articles/Factory_Reset/)** will definitely wipe out all Android stuff and unfortunately much more. Hard work will be needed to set up the device for pure Sailfish OS use after the reset.
```
vigejolla, line 63: That link doesn't look right

Support/Help_Articles/Android_App_Support/Removing_Android_App_Support/README.md
```diff
@@ -0,0 +1,65 @@
...
+</div>
+
+
+3) Restart the device.
+
+_This is the recommended way. It is expected to work correctly on Sailfish 4 releases and thereafter._
+
+# Brutal way
+
+Consider using this method if the [Clean removal](/Support/Help_Articles/Android_App_Support/Removing_Android_App_Support/#clean-removal) fails to work, for some reason.
```
vigejolla, line 42: All the other links were in bold, was there a reason why this one isn't?

vigejolla:
I think the rearranging part looks nice now. The two commits related to that could be squashed together though. And those two comments on the "removing" page are still valid.

Support/Help_Articles/Android_App_Support/Removing_Android_App_Support/README.md
jovirkku: Fixed now.

Support/Help_Articles/Android_App_Support/Removing_Android_App_Support/README.md
jovirkku: Hmm I wonder what happened. Fixed now.

vigejolla (approved):

PR #205, opened 2022-12-12
Update APKPure download instructions

vigejolla:
Downloading APK files from APKPure does not work on our browser with a simple click. Opening the link in new tab works.


jpwalden (approved):
Thanks, good update, this way the download works.

jovirkku (approved):
It works!

rainemak (approved):
There's an issue behind this as well. When tapping green-and-white "Download" the url starts with apkpure://install-app? which is not supported. Please file an another issue about this if not reported already.

From article point of view, this is still better than what it used to be.

PR #211, opened 2022-12-14
Document how to get backtrace with gdb at boot from any service

LaakkonenJussi:
Added documentation about how to get backtrace from services that crash at boot. This requires only gdb and the relevant debuginfo packages to be installed and shows straightforward way for service debug with gdb.

### DIFF
--- a/Reference/Sailfish_OS_Cheat_Sheet/README.md
+++ b/Reference/Sailfish_OS_Cheat_Sheet/README.md
@@ -326,6 +326,58 @@ Enabled=0
 Enabled=0
 ```
 
+### System service debugging
+
+Sometimes changes to system services that are being started at boot cause unwanted results and the service crashes. Or the service crashes only on device boot. These can be resolved by checking the backtraces of such service. This set of instructions applies generally for other services as well.
+
+This requires only `gdb` from the `sdk` repository:
+```nosh
+ssu ar sdk && pkcon refresh && pkcon install gdb
+```
+
+And the relevant debuginfo (and debugsource, installed as dependency) packages to have meaningful output.
+
+In order to get the backtrace out from a service during boot create a following gdb command file (suffix does not matter):
+```ini
+set width 0
+set height 0
+set verbose off
+set logging file <path_to_log_file>
+set logging on
+handle SIG33 pass nostop noprint
+run
+backtrace full
+```
+
+And add a systemd drop-in file to `/usr/lib/systemd/system/<service_name>.service.d/override.conf` with the content to override the existing service file variables:
+```ini
+[Service]
+ExecStart=
+ExecStart=/usr/bin/gdb --batch --command=<command_file> --args <copy_this_part_from_original_service_file>
+Restart=no
+```
+
+`ExecStart` needs to be cleared to avoid error `Service has more than one ExecStart= setting, which is only allowed for Type=oneshot services. Refusing`. `Restart` is prevented in order to avoid overwriting the log file.
+
+For example with ofono following can be used. Put the following (change path to `--command` if needed) and do:
+```nosh
+mkdir -p /usr/lib/systemd/system/ofono.service.d/
+cat << EOM >/usr/lib/systemd/system/ofono.service.d/override.conf
+[Service]
+ExecStart=
+ExecStart=/usr/bin/gdb --batch --command=/root/ofono.gdb --args /usr/sbin/ofonod -n --nobacktrace \$OFONO_ARGS \$OFONO_DEBUG
+Restart=no
+EOM
+```
+
+To make sure that the change is applied do:
+```nosh
+systemctl daemon-reload
+```
+And reboot the device. The log is available with the backtrace in `<path_to_log_file>` pointed in the `<command_file>`.
+
+When there is no need for gdb anymore the `override.conf` can be removed. After removal `daemon-reload` and service restart is required for the change to be in effect.
+
 ## Restart System Services
 
 Restart user session


vigejolla (approved):
LGTM

Reference/Sailfish_OS_Cheat_Sheet/README.md
```diff
@@ -326,6 +326,54 @@ Enabled=0
...
+set height 0
+set verbose off
+set logging file <path_to_log_file>
+set logging on
+handle SIG33 pass nostop noprint
+run
+backtrace full
+```
+
+And add a systemd drop-in file to `/usr/lib/systemd/system/<service_name>.service.d/override.conf` with the content to override the existing service file variables:
```
Thaodan, line 352: If using devel-su -c 'systemctl edit <service>' you can avoid the manual steps to create the directory, the files and system daemon reload.

Reference/Sailfish_OS_Cheat_Sheet/README.md
```diff
@@ -326,6 +326,54 @@ Enabled=0
...
 
+### System service debugging
+
+Sometimes changes to system services that are being started at boot cause unwanted results and the service crashes. Or the service crashes only on device boot. These can be resolved by checking the backtraces of such service. This set of instructions applies generally for other services as well.
+
+This requires only `gdb` from the `tools` repository:
+```ssu ar tools
+pkcon refresh
+pkcon install gdb
+```
```
Thaodan, line 337: You can use `pkcon refresh && pkcon install gdb` for easier refresh.

Thaodan (commented):
All in all looks good to me, only some smaller suggestions.
I'm wondering is there a bug for this?

Reference/Sailfish_OS_Cheat_Sheet/README.md
LaakkonenJussi: I need to try this. I was not aware of that, I usually do the things the hard way.

Reference/Sailfish_OS_Cheat_Sheet/README.md
LaakkonenJussi: I'd rather use `zypper ref tools` but I guess `pkcon` is preferred usually as it is installed already on the devices.

Reference/Sailfish_OS_Cheat_Sheet/README.md
```diff
@@ -326,6 +326,54 @@ Enabled=0
 Enabled=0
 ```
 
+### System service debugging
+
+Sometimes changes to system services that are being started at boot cause unwanted results and the service crashes. Or the service crashes only on device boot. These can be resolved by checking the backtraces of such service. This set of instructions applies generally for other services as well.
+
+This requires only `gdb` from the `tools` repository:
+```ssu ar tools
```
vigejolla, line 334: I just now noticed that you had the first line of commands on the same line as backticks. This does not work, they need to be on separate lines. The same error is on all instances where you added code blocks.

And then I noticed that you used the tools repo, which I believe doesn't work on release devices. I think gdb is found on sdk repo there. I don't have any release device at hand so I can't easily verify though.

vigejolla (changes requested):

Reference/Sailfish_OS_Cheat_Sheet/README.md
LaakkonenJussi: Actually, `systemctl edit` creates the overrides to `/etc/systemd/system/` and then the changes made to the location we use by default (`/usr/lib/systemd/system/`) do not work afterwards for the service. Apparently one would have to manually remove the `/etc/systemd/system/<servicename>.service.d/*.conf`. I think manual steps here are more obvious. Usually systemd tools just obfuscate things (I do not trust them).

Reference/Sailfish_OS_Cheat_Sheet/README.md
LaakkonenJussi: Changed to oneliner.

Reference/Sailfish_OS_Cheat_Sheet/README.md
LaakkonenJussi: Also, when and if we change the location where `systemctl edit` makes the changes we'd have then outdated location here in the docs. By telling the exact location here and forcing to do a bit of manual labor the details are always permanent. This isn't really something one needs to do even on monthly, let alone weekly basis.

Reference/Sailfish_OS_Cheat_Sheet/README.md
LaakkonenJussi: Right, I guess the format details were not correct. Made the changes to format.

And changed to use the sdk repo. Although on devel devices it does not seem to have `gdb`..

LaakkonenJussi: Oh, no, apparently `gdb` is in sdk repo. Nevermind that.

LaakkonenJussi:
> LGTM, but please squash the commits

Will do, always :).

Reference/Sailfish_OS_Cheat_Sheet/README.md
Thaodan: > Actually, `systemctl edit` creates the overrides to `/etc/systemd/system/` and then the changes made to the location we use by default (`/usr/lib/systemd/system/`) do not work afterwards for the service. Apparently one would have to manually remove the `/etc/systemd/system/<servicename>.service.d/*.conf`. I think manual steps here are more obvious. Usually systemd tools just obfuscate things (I do not trust them).

the changes can be revert with systemctl revert.

LaakkonenJussi: We could separately raise this issue in the documentation with the warning that `revert` will cancel all changes and the changes are stored elsewhere than what we commonly use as location when `edit` is used. Perhaps a reference later on then to that section somewhere in this added documentation is in place. Therefore, resolving this.

Thaodan: > We could separately raise this issue in the documentation with the warning that `revert` will cancel all changes and the changes are stored elsewhere than what we commonly use as location when `edit` is used. Perhaps a reference later on then to that section somewhere in this added documentation is in place. Therefore, resolving this.

I wasn't aware of that, will check.

PR #214, opened 2022-12-16
Replace YAML instructions with SPEC instructions

vigejolla:


martyone:
It's strange the diff is so big - after a bit closer glance it seems there may not actually be that much changes, it just fails to match the old and new content. It might help to preserve the section titles or move their changes into a separate commit - could you please try that? There may be other changes that, when split into a separate commit, would make the diff much smaller.

vigejolla:
> It's strange the diff is so big - after a bit closer glance it seems there may not actually be that much changes, it just fails to match the old and new content. It might help to preserve the section titles or move their changes into a separate commit - could you please try that? There may be other changes that, when split into a separate commit, would make the diff much smaller.

Yeah, the section headings have changed, as they are really the keywords used in the files - spec files have different keywords than yaml files. The other thing which affects the large looking diff is the order of paragraps, as I had to move things around a bit.

I'll try to separate these into separate commits, let's see...

vigejolla:
It should be a bit easier to review now. Please note that this splitting of the commit was done solely for the purpose of the review, I intend to squash the commits before merging.

martyone:
hm, did it really go as intended? To me it seems the first commit does some moves that the second commit reverts.

I would definitely not squash the commits - atomic commits are very valuable.

vigejolla:
:facepalm: let me try again...

martyone (approved):
Looks better now.

Wouldn't it be more natural to list BuildRequires before Requires? I see that our spectacle does it the opposite way but e.g. here https://rpm-packaging-guide.github.io/ you can see BuildRequires first. This would save us from some unnecessary shuffling of the sections too. Leave the Description on its original place as well and all we need to move is the "About the PkgConfigBR and PkgBR Keywords" section :)

It would be nice to have a different type of arrow from .spec to .rpm in the diagram. The relation is quite different from the one between .pro and .spec.

Please do NOT squash the commits.

vigejolla:
> Wouldn't it be more natural to list BuildRequires before Requires? I see that our spectacle does it the opposite way but e.g. here https://rpm-packaging-guide.github.io/ you can see BuildRequires first. This would save us from some unnecessary shuffling of the sections too. Leave the Description on its original place as well and all we need to move is the "About the PkgConfigBR and PkgBR Keywords" section :)

It would. We don't use spectacle anymore, so we could change the order in the templates as well. That's a minor detail, but I'll file another task about that. And change the order in the document already in this PR.

> It would be nice to have a different type of arrow from .spec to .rpm in the diagram. The relation is quite different from the one between .pro and .spec.

I agree. I just tried to mimic the original diagram. I'll change the arrow type.

> Please do NOT squash the commits.

Ok, ok :)

PR #215, opened 2022-12-16
Create Camera document

jpwalden:


Support/Help_Articles/Camera/README.md
```diff
@@ -0,0 +1,87 @@
...
+
+The picture below presents the viewfinder view of the camera. On top there is the toolbar row, and by touching it all of the options will appear.
+
+At bottom left you can adjust the exposure easily. The capture button is at bottom in the middle in this case.
+
+The circle in the middle is the focus area. The camera sets the focus according to the distance and illumination of it. It is possible to move the focus area to a different location in the "Tap to focus" mode (see "Adjusting settings" below).
+
+<div class="flex-images" markdown="1">
+
+* <a href="viewfinder.png" class="narrow-image"><img src="viewfinder.png" alt="viewfinder"></a>
```
jovirkku, line 37: This pic is from a Jolla Phone. The viewfinders of the recent Xperias have the choice of camera buttons, too, indicated with numbers 1.0, 2.0 and 0.6 (representing the normal, tele and wide-angle lenses). As those may be a bit unclear to some readers it would be good to renew the pic and explain those buttons.

Support/Help_Articles/Camera/README.md
```diff
@@ -0,0 +1,87 @@
...
+
+The settings of Jolla Camera are listed below. Please read more about the settings in [this article](https://jolla.zendesk.com/hc/en-us/articles/201440317).
+
+#### Settings useful at the moment of capturing a picture
+
+* Colour filtering on/off
+* Capture delay:  0,  3,  10 or  15 seconds
+* Flash light: Automatic , Flash off , Flash on 
+* Viewfinder grid: No grid/Thirds grid 
+* White balance can be adjusted from the right side of the view finder Automatic (AWB)
```
jovirkku, line 66: In addition to the white balance, one can increase or decrease the exposure with the vertical slider.

Support/Help_Articles/Camera/README.md
```diff
@@ -0,0 +1,87 @@
...
+
+### Adjusting settings
+
+The settings of Jolla Camera are listed below. Please read more about the settings in [this article](https://jolla.zendesk.com/hc/en-us/articles/201440317).
+
+#### Settings useful at the moment of capturing a picture
+
+* Colour filtering on/off
+* Capture delay:  0,  3,  10 or  15 seconds
+* Flash light: Automatic , Flash off , Flash on
```
jovirkku, line 64: Between the flash and viewfinder grid buttons there is the "A" which has the text "Automatc exposure". That needs some clarification, in particular, as there does not seem to be an option for "manual exposure" - other than using the vertical exposure slider. But this "A" button does not seem to do anything useful. It is obscure to me, so it might be that for the camera users, too.

Support/Help_Articles/Camera/README.md
```diff
@@ -0,0 +1,87 @@
...
+### Shutter button
+
+The location of the capture button can be moved on the screen. Tap and hold the viewfinder screen until several circles appear. Tap one to move the capture button there.
+
+### Adjusting settings
+
+The settings of Jolla Camera are listed below. Please read more about the settings in [this article](https://jolla.zendesk.com/hc/en-us/articles/201440317).
+
+#### Settings useful at the moment of capturing a picture
+
```
jovirkku, line 61: These settings are brought to the display by touching the top of the viewfinder having the auto-flash and A symbols.

Support/Help_Articles/Camera/README.md
```diff
@@ -0,0 +1,87 @@
...
+* Colour filtering on/off
+* Capture delay:  0,  3,  10 or  15 seconds
+* Flash light: Automatic , Flash off , Flash on 
+* Viewfinder grid: No grid/Thirds grid 
+* White balance can be adjusted from the right side of the view finder Automatic (AWB)
+* Camera modes: Still camera  and Video camera 
+* Camera selection:  Main camera,  Front camera.
+
```
jovirkku, line 69: When the settings buttons are visible, there is also the "reset to defaults" button at the bottom right corner of the viewfinder. This button brings the capture delay and the camera flash settings to their defaults (but surprisingly, does not do it for the exposure). Hmm. A bug or not?

Support/Help_Articles/Camera/README.md
jovirkku, line 64: We should not call it the "flash light" as this (according to a dictionary) means a "pocket lamp". It is either "flash" or "camera flash".  I checked this with David L.J.

Support/Help_Articles/Camera/README.md
```diff
@@ -0,0 +1,87 @@
...
+<div class="flex-images" markdown="1">
+
+* <a href="icon_camera.png" class="narrow-image"><img src="icon_camera.png" alt="icon_camera"></a>
+  <span class="md_figcaption">
+  </span>
+</div>
+
+Two ways to take a picture once the camera app is opened:
+
```
jovirkku, line 21: There is even a 3rd way: by pushing the lock screen up. This is possible if the "Settings > System > Gestures" has the (default) setting "Quick access to Camera" enabled.

Support/Help_Articles/Camera/README.md
jpwalden: This feature is not clear for me either

Support/Help_Articles/Camera/README.md
jpwalden: Good point, it is good to explain that as well, I'll add that.

Support/Help_Articles/Camera/README.md
jpwalden: Good point, it is good to explain that as well, I'll add that.

Support/Help_Articles/Camera/README.md
jpwalden: Yes, I just used the same wording, but that was not the best wording on that. I'll change that

Support/Help_Articles/Camera/README.md
jpwalden: Yes we should, I'll add that.

Support/Help_Articles/Camera/README.md
jpwalden: Thanks, Added second pic from X10 II and additional explanations.

Support/Help_Articles/Camera/README.md
jpwalden: added

Support/Help_Articles/Camera/README.md
jpwalden: Added

Support/Help_Articles/Camera/README.md
jpwalden: Added

Support/Help_Articles/Camera/README.md
jpwalden: updated

Support/Help_Articles/Camera/README.md
jpwalden: Added

Support/Help_Articles/Camera/README.md
vigejolla: The second pic is from Xperia 10 II, but the text says III. By default Xperia 10 III does not have the lense buttons.

jpwalden: Yes, typo -thanks for noticing it. Fixed

Support/Help_Articles/Camera/README.md
vigejolla: On some devices (e.g. Jolla C) you can select various (automatic) exposure modes. On some devices (e.g. Jolla Tablet) it seems automatic exposure is not available at all, and the button is missing. On my Xperia 10 III there is the "A" button but no others, so I believe there is automatic exposure which can't be turned off.

vigejolla (approved):

PR #217, opened 2022-12-20
Describe Target device filesystem image in more detail

vigejolla:


Reference/Toolchain/README.md
```diff
@@ -28,6 +28,8 @@ Scratchbox 2 achieves this by employing target CPU emulation and by composing a
 
   - Target device filesystem image
       - Target-compatible binaries
+      - Target device filesystem image is needed for each architecture (armv7hl, aarch64, i486)
+      - Target device filesystem image becomes hardware adaptation specific once hardware specific repositories are added or otherwise installed
```
martyone, line 32: I am afraid one cannot really understand what we mean here unless we expand more on the nuances between the terms we use here. How about to just add single line "One for each target CPU architecture (armv7hl, aarch64, i486)" and avoid mentioning hardware adaptation here? It's not just HA level that may make a build target specific to a particular device anyway. It may be a much higher level choice of packages as well.

vigejolla: It's a difficult balance, trying to explain the terms so that they are understandable, while at the same time trying to refrain from going into too much detail. I think we do need to explain that sometimes there is a need to have device specific targets, and those are created from the generic architecture specific targets by adding or replacing packages and/or repositories. At the same time we should try to convey the idea that for most users this is not necessary and the generic targets are enough.

martyone: I would mention the need for somehow specific targets only where the related use case is described. E.g. in a HA related how-to or so. And keep just the generic one target per CPU arch rule here.

rainemak: "One for each target CPU architecture (armv7hl, aarch64, i486)" sounds good. Let's double check what does HADK pdf & docs pages say.

vigejolla: HADK refers to https://docs.sailfishos.org/Tools/Platform_SDK/Target_Installation/ which has the chapter https://docs.sailfishos.org/Tools/Platform_SDK/Target_Installation/#separate-targets in the end. That describes the device-specific targets in more detail.

rainemak (approved):
Cheers Ville!

PR #219, opened 2022-12-21
Gallery app

jovirkku:


jpwalden (commented):
Looks good to me, however should the nav order be 485 since that is between 470 and 500?

jovirkku:
> should the nav order be 485 since that is between 470 and 500?

The previous doc is "Flashing... (470) and the next is "Importing..." (500). 
Gal is closer to Fla than to Imp - and H-docs are completely missing. This is why I chose an index closer to the previous as there won't probably be many docs between "Flashing" and "Gallery". More space to add H docs.

jpwalden:
> > should the nav order be 485 since that is between 470 and 500?
> 
> The previous doc is "Flashing... (470) and the next is "Importing..." (500). Gal is closer to Fla than to Imp - and H-docs are completely missing. This is why I chose an index closer to the previous as there won't probably be many docs between "Flashing" and "Gallery". More space to add H docs.

ok, makes sense.

jpwalden (approved):
Looks good to me.

vigejolla (approved):

PR #220, opened 2022-12-21
Lock jekyll-sass-converter to version 2.2.0

vigejolla:
Version 3.0.0 breaks the build tests in our CI


sledges (approved):
Source branch has a green tick https://github.com/sailfishos/docs.sailfishos.org/commits/jb59688
So here's one more:) :heavy_check_mark:

PR #223, opened 2022-12-29
Locations of apps in the filesystem

jovirkku:


jovirkku:
This document is somewhat concise. The justification for it is that quite often the customers ask questions about the app data locations in Jolla Care. They want to save some valuable data by using the command line when nothing else helps.

jpwalden (commented):
Looks good to me, however maybe this link could be added same way as other links, so that it is embedded.
This directory contains SQL database ("commhistory.db*") of text messages. See https://jolla.zendesk.com/hc/en-us/articles/203569178 .

jovirkku:
> Looks good to me, however maybe this link could be added same way as other links, so that it is embedded. This directory contains SQL database ("commhistory.db*") of text messages. See https://jolla.zendesk.com/hc/en-us/articles/203569178 .

I have fixed this.

Support/Help_Articles/Locations_of_Apps/README.md
```diff
@@ -0,0 +1,154 @@
...
+  
+
+## Harbour apps
+
+Harbour apps are available in the Jolla Store app. They have been developed by the Sailfish community. The majority of Harbour apps (there may be deviations) keep their data in this directory tree:
+```
+$MYHOME/.local/share/
+```
+
+There should be a dedicated subdirectory beginning with "harbour-" for each Harbour app.
```
vigejolla, line 141: This is outdated information. These days the subdirectory does not necessarily begin with "harbour-". It may though. Making matters worse, many apps have migrated away from old "harbour-" directories, leaving the old directories behind. So for a given app, there might be the old directory, say "harbour-myapp", which contains old files which are not used anymore, and then there is the new directory, say "org.developer/myapp", which contains current files. I don't know how much into detail we want to go here?

jovirkku: Right. I need to reconsider the text. Secondly, I happened to find some more useful text for this document.

jovirkku: I decided to remove the Harbour chapter as it is a bit vague. It was not really about "where they keep their data".

vigejolla: I suppose that's a good decision at least for the time being. We might want to consider writing something useful about them in the future.

vigejolla (approved):

PR #235, opened 2023-01-16
Add a page for using and registering file/url handlers. JB#59454

pvuorela:
Add some words about file type and url handlers. These sort of fall between the cracks of modules and having now some new capabilities here.


vigejolla (approved):
Almost LGTM. However, the new page could be added to https://docs.sailfishos.org/Reference/Core_Areas_and_APIs/Apps_and_MW/ as well. And really, there's no reason why the list on that page is even created manually when it could be created automatically, see e.g. https://docs.sailfishos.org/Support/Help_Articles/. But that's out of scope for this task -> approving as is.

pvuorela:
Added now to the other page too.

vigejolla (approved):

PR #237, opened 2023-01-17
Use where for filtering Apps and MW subpages

vigejolla:
Original code failed on Array comparison while sorting pges, most likely because array had pages with titles which could be interpreted as integers. Filtering the pages before sorting them fixes the issue, and also makes the filtering inside for loop redundant.


vigejolla:
Note: I'm not 100% sure this fixes the issue, as the issue does not manifest itself on a local build, or even on the CI build. The only way to test is to merge and see what happens. We should at some point fix our CI so that it would be more close to the real thing so it would catch this kind of issues before merging.

pvuorela (approved):

PR #241, opened 2023-01-19
Publishing Sailfish OS updates

jovirkku:


jpwalden (changes requested):
"We have a small number of technically oriented people, "geeks", who have signed a Non-Disclosure Agreement (NDA) with Jolla."
I would edit this a bit since, I wouldn't want to call cbeta people as "geeks" since some of them might not like that.  I think the "technically oriented people" is good, so I would remove the geeks.
Then a comment, I'm not sure how actively we are to get more people on the cbeta group (I would assume we are, at least I am :) ), but would be good to have here some information how you can apply to the cbeta group if you are interested. This is just a comment since I'm not aware how actively we are looking more people and I'm also not sure how the process even works.

jovirkku:
Everything about cbeta and community testing was removed.
Added the link to the table of official releases.

jpwalden (approved):
Looks good to me

Support/Help_Articles/Publishing_Sailfish_OS_Updates/README.md
```diff
@@ -0,0 +1,50 @@
+---
+title: Publishing Sailfish OS Updates
+permalink: Support/Help_Articles/Publishing_Sailfish_OS_Updates/
+parent: Help Articles
+layout: default
+nav_order: 625
```
vigejolla, line 6: Now this page is after "QR Code Reader", shouldn't it be before?

jovirkku: Yes, definitely.
Fixed.

vigejolla (approved):

PR #248, opened 2023-01-24
Add good quality icon

vigejolla:


sledges:
Is it intentional that this doesn't look the same as https://raw.githubusercontent.com/sailfishos/docs.sailfishos.org/master/favicon.ico ?
I like that it is square though, I wish I made the favicon.ico also square (still blue however, to match the top-left blue SFOS logo in the actual page).

vigejolla:
> Is it intentional that this doesn't look the same as https://raw.githubusercontent.com/sailfishos/docs.sailfishos.org/master/favicon.ico ? I like that it is square though, I wish I made the favicon.ico also square (still blue however), to match the top-left blue SFOS logo in the actual page.

Kind of intentional. I didn't want to change the existing favicons, because they already look nice, and in my opinion better on my firefox tabs. On the other hand, it was explicitly mentioned in task that the icons from sailfishos.org were to be used so I used that for apple-touch-icon, which is what our browser uses.

sledges (approved):

PR #249, opened 2023-01-25
Corrections to i18n docs

pvuorela:
Bokmål and Sanskrit are languages, not scripts. Removed those and added the ones we have font support for. All except thai and arabic have also input support.

Removed also reference to Nuance being which hinted it being a Maliit plugin. That's incorrect, the jolla-keyboard is a maliit plugin and jolla-keyboard internally has pluggable support for using Nuange engine.

Signed-off-by: Pekka Vuorela <pekka.vuorela@jolla.com>

### DIFF
--- a/Reference/I18n/README.md
+++ b/Reference/I18n/README.md
@@ -36,7 +36,7 @@ The translation file will also define the layout directionality (e.g.: left-to-r
 
 Different locales require different input methods. In some locales, a QWERTY-layout virtual keyboard is the most common input method, but other locales require different keyboard layouts, and in some locales the most common input method is hand-writing-recognition.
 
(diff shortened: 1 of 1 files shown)


sledges (approved):
LGTM, only commit description text:

> Removed also reference to Nuance being which hinted it being a Maliit plugin

I believe the first "being" should be dropped.

> for using Nuange engine.

Is Nuange -- a Nuance's engine, or a typo? The only place I found meniton of both is:..
https://archive.org/stream/Personal_Computer_World_2009-06_VNU_Business_Publications_GB/Personal_Computer_World_2009-06_VNU_Business_Publications_GB_djvu.txt

pvuorela:
Yep, latter Nuange was a typo. Adjusted that and the other mentioned bad wording.

PR #252, opened 2023-01-29
Link to forum wiki instead of mer wiki

b100dian:
Since the merproject.org/wiki is down for some time now, change this link to forum too..


vigejolla (approved):
Awesome, thanks!

PR #255, opened 2023-02-03
Add Firewall description from ZenDesk

LaakkonenJussi:
Moved https://jolla.zendesk.com/hc/en-us/articles/360017800813-Firewall-in-Sailfish-OS to forum format. Also updated any outdated or wrong info on the page.

### DIFF
--- a/Support/Help_Articles/Firewall/README.md
+++ b/Support/Help_Articles/Firewall/README.md
@@ -1,0 +1,338 @@
+---
+title: Firewall 
+permalink: Support/Help_Articles/Firewall/
+parent: Help Articles
+layout: default
+nav_order: 460
+---
+
+Firewall in Sailfish OS is controlled by ConnMan by its internal rules and configuration files. A proper firewall was introduced first in Sailfish OS update 3.0.1.
+
+A normal Sailfish OS user should not need to read this document. It is intended for developers and for Internet enthusiasts.
+
+## Basics
+
+ConnMan uses iptables to manage the firewall. This is communicated using unix sockets. Firewall is set up when ConnMan starts and cleared when it stops.
+
+
+## Configuration files
+
+The iptables firewall configurations are supported as of ConnMan 1.32+git41 (Sailfish OS update 3.0.1). These configurations are stored in the format described in this document in the following files and locations:
+
+  - `/etc/connman/firewall.conf`
+  - `/etc/connman/firewall.d/*-firewall.con`
+
+1) The main configuration
+2) A directory in which all files ending with `-firewall.conf` are used in alphabetical order.
+
+
+## Main configuration of Sailfish OS firewall
+
+The default configurations are delivered in package **configs-connman-sailfish**.
+
+
+### Rules
+
+The package **configs-connman-sailfish** sets the following rules in the following files:
+
+#### /etc/connman/firewall.conf
+
+  - IPv4
+    - Allow established and related packages (`-m conntrack --ctstate RELATED,ESTABLISHED`)
+    - Allow all incoming packages to loopback interface (required by connman DNS resolving)
+    - Default: block all incoming traffic by default (filter table INPUT chain default policy)
+  - IPv6
+    - Allow established and related packages (`-m conntrack --ctstate RELATED,ESTABLISHED`)
+    - Allow all incoming packages to loopback interface (required by connman DNS resolving also for IPv6)
+    - Default: block all incoming traffic by default (filter table `INPUT` chain default policy)
+
+
+#### /etc/connman/firewall.d/10-allow-dhcp-firewall.conf
+
+  - IPv4
+    - Workaround possible issue with **DHCPv4** when ACK packet is not determined to be in `RELATED` state by iptables during allocation or lease.
+    - Allows traffic to **UDP port 68** that is open only during processing DHCP requests
+  - IPv6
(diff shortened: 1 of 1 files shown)


Support/Help_Articles/Firewall/README.md
```diff
@@ -0,0 +1,337 @@
...
+     ```
+     and IPv6:
+     ```
+     watch -n 1 ip6tables -t filter -L -v -n 
+     ```
+     Use Ctrl+C to quit.
+  3. Check the output and amount of dropped packets in Chain INPUT (policy DROP).
+  4. Run the application and check which counters increase, the topmost DROP will increase if packets are being dropped, the packet counters on each rule will increase if packets match that rule
+  5. If the problem is with dropping packets, you can try:
+     5.1 a temporary solution to allow all traffic by default on IPv4:
```
vigejolla, line 317: Markdown doesn't really support having "five point something" in numbered lists. My suggestion: Use a bullet list as a second list, so something like:
```
  5. If the problem is with dropping packets, you can try either:
   * a temporary solution to allow all traffic by default on IPv4:
     or
   * if you wish it to be permanent, add file **/etc/connman/firewall.d/99-accept-all-firewall.conf** and write the following lines to it:
```

Support/Help_Articles/Firewall/README.md
```diff
@@ -0,0 +1,337 @@
...
+     IPv4.INPUT.POLICY = ACCEPT
+     IPv6.INPUT.POLICY = ACCEPT
+     ```
+     and restart connman with
+     ```
+     systemctl restart connman
+     ```
+  6. If the problem was solved with this, please report this problem at forum.sailfishos.org.
+
+If the device for some reason does not boot, enter the recovery mode and setup the same content for the configuration file as in section 5.2. If the device boots after this please report this problem with logs at forum.sailfishos.org.
```
vigejolla, line 337: That "section 5.2" should probably be a link to somewhere

Support/Help_Articles/Firewall/README.md
```diff
@@ -0,0 +1,337 @@
...
+The rest of the configurations are loaded from `/etc/connman/firewall.d` in alphabetical order.
+
+### Processing order of rules:
+  - Since the keys from files are loaded in alphabetical order, the rules in file with, e.g., 00 prefix will be processed and added first.
+  - Rules from main config are added as last rules into iptables. These are considered as base rules.
+  - When dynamic rules are enabled these are inserted to top in iptables
+  - As for example:
+    - Configs and processing order:;
+      1. base `firewall.conf` - has `[General]` rules and POLICY set
+      2: `firewall.d/10-firewall.conf` - has `[wifi]` rules
```
vigejolla, line 153: These numbers 2-4 should also have '.' after the number instead of ':'

vigejolla (changes requested):
I'm sure the content itself is fine, but those lists need a little bit of finetuning

Support/Help_Articles/Firewall/README.md
LaakkonenJussi: Yeah, I did a reformat on this part. Thanks for pointing it out. I hope it is now clearer even. No need to mention boot issues I guess, since that has not been the case for years.

Support/Help_Articles/Firewall/README.md
LaakkonenJussi: Yeap, missed that. Thanks.

Support/Help_Articles/Firewall/README.md
```diff
@@ -332,6 +332,6 @@ If, for some reason, there is a problem with some application that would seem li
      ```
      systemctl restart connman
      ```
-  6. If the problem was solved with this, please report this problem at forum.sailfishos.org.
+  6. If the problem was solved with this, please report this problem at [SailfishOS forum](forum.sailfishos.org).
```
vigejolla, line 335:
```suggestion
  6. If the problem was solved with this, please report this problem at [Sailfish OS Forum](https://forum.sailfishos.org).
```

Support/Help_Articles/Firewall/README.md
```diff
@@ -332,6 +332,6 @@ If, for some reason, there is a problem with some application that would seem li
...
 
-If the device for some reason does not boot, enter the recovery mode and setup the same content for the configuration file as in section 5.2. If the device boots after this please report this problem with logs at forum.sailfishos.org.
+If there are other network related issues with the device try the example configuration above (**99-accept-all-firewall.conf**) to solve the issue if it seems like a firewall issue. If this solves any issue please report the problem with logs at [SailfishOS forum](forum.sailfishos.org).
```
vigejolla, line 337:
```suggestion
If there are other network related issues with the device try the example configuration above (**99-accept-all-firewall.conf**) to solve the issue if it seems like a firewall issue. If this solves any issue please report the problem with logs at [Sailfish OS Forum](https://forum.sailfishos.org).
```

vigejolla (changes requested):

vigejolla (approved):
LGTM, one nitpick still though: There should be a space between "Sailfish" and "OS" in "Sailfish OS Forum"

jpwalden (approved):
Looks good to me

LaakkonenJussi:
> LGTM, one nitpick still though: There should be a space between "Sailfish" and "OS" in "Sailfish OS Forum"

Ah yeah, minor issue but `sed`able ;). Done.

PR #258, opened 2023-02-08
Update the script 'Collect persistent logs'

jovirkku:


jovirkku:
Update the script from v.3 to v.4. 
This is to ensure that the script matches AppSupport v.11.

It is known that lxc-attach, when running the script, prints out some warnings on 4.5.0 while on 4.6.0 it does not. It works correctly, though.

The author of the script is Juho. The script is actually kept in GitHub. Our document has it attached. Should we use a direct link rather than updating the script here every now and then?

jpwalden (approved):
As far as I can understood this is ok.

jusa:
If I knew the script was here I would have updated it directly here as well.. :) So ok for me to have it here, in the future I will modify this one directly.

jusa (approved):

PR #261, opened 2023-02-13
Improve gather logs script.

jusa:


mkosola (approved):

jovirkku:
On 4.5.0, a few failure messages are shown from lxc-attach. But this is known to get better on 4.6.0 (a new version of lxc-attach) and this seems to be the case - I tested.

On 4.6.0, the following error message appears: "Error dumping service into status_t: 1 (Unknown error -1) wifin180211

The list of installed packages would be more convenient to read if it was in alphabetical order.

The script seems to work correctly.

jovirkku (approved):
https://github.com/sailfishos/docs.sailfishos.org/pull/261#issuecomment-1427774805

jusa:
> On 4.5.0, a few failure messages are shown from lxc-attach. But this is known to get better on 4.6.0 (a new version of lxc-attach) and this seems to be the case - I tested.
> 
> On 4.6.0, the following error message appears: _"Error dumping service into status_t: 1 (Unknown error -1) wifin180211_
> 
> The list of installed packages would be more convenient to read if it was in alphabetical order.
> 
> The script seems to work correctly. The logs look like correct.

I modified it so that the errors are directed to files instead of being user-visible.

mkosola (approved):

PR #264, opened 2023-02-16
Add alphanumeric code jb60154

jovirkku:


vigejolla:
I believe those [Add browser reset. Fix hyperlinks.](https://github.com/sailfishos/docs.sailfishos.org/pull/264/commits/8fb21beb90c3ce6d37a8121b5800223ec238b627) commits should already be on master. Please rebase on top of master.

Support/Help_Articles/Device_Lock_and_Security_Code/README.md
```diff
@@ -44,13 +44,44 @@ On a shared device (having multiple user accounts), there is no Device lock item
...
+0123456789
+,.!?;:&%#()=+- and space
+```
 
-Please note that this same code is also used for the encryption of user data.
+If a wrong character is typed (which is possible as the standard keypad is used) the warning "Disallowed characters" immediately appears on the display.
 
-Support for an alphanumeric security code is coming up in one of the releases after 4.4.0.
+The minimum length of the code is 5 digits but we recommend a considerably longer code. The maximum is 42.
+By default, the security code dialog has the numeric keypad (see Pic 2 above). The alphanumeric keypad can be exposed by tapping the button at the top left corner of
```
vigejolla, line 58: To make future reviews easier, I recommend that you remove the line break after "of".

vigejolla (changes requested):
Please rebase and if possible, remove that one line break. The line break doesn't matter now, but in the future it might as someone editing the text might also move the line break, which makes reviewing difficult.

vigejolla:
Those extra commits are still there. Please do as follows:
1. execute `git rebase -i dff6930fc73bd2e4cb648f708930c91fed3c8095`
2. put `d` in front of the extra commits
3. after finishing with the above rebase, run `git rebase master`
4. `git push -f`

vigejolla (approved):

PR #278, opened 2023-03-08
Update README.md

vigejolla:
At least on my devices, `reboot` needs root privileges, so it does not actually reboot the device if executed after `exit`.


jovirkku (approved):
Yes, a good catch. My X10iii does not reboot, either, without the root permissions.

rainemak (approved):

PR #294, opened 2023-03-17
Fix permalink

vigejolla:


vigejolla:
We should try to catch this kind of errors in the CI :roll_eyes:

pvuorela (approved):

PR #296, opened 2023-03-20
Add release 4.5.0.19 to the table of releases

jovirkku:


jovirkku:
This PR adds the latest 4.5.0 hotfix to the table of releases.

A concern: The table has the latest ID (4.5.0.19) but the date of the 1st 4.5.0 release (4.5.0.16). 
Do we want to 
a) change the date to 2023-03-20, or
b) shorten the ID to 4.5.0 and show the date of the latest hotfix
c) list all 4.5.0 releases separately
d) something else, what?
Personally, I would select option b).

vigejolla (approved):

vigejolla:
> This PR adds the latest 4.5.0 hotfix to the table of releases.
> 
> A concern: The table has the latest ID (4.5.0.19) but the date of the 1st 4.5.0 release (4.5.0.16). Do we want to a) change the date to 2023-03-20, or b) shorten the ID to 4.5.0 and show the date of the latest hotfix c) list all 4.5.0 releases separately d) something else, what? Personally, I would select option b).

I think the option b is the worst, because the information it provides is clearly wrong. If we show the ID as "4.5.0", the release date should be the release date of 4.5.0 and not some hotfix. But anyway, merging this as is now, let's continue this discussion elsewhere.

PR #298, opened 2023-03-22
Document Buteo profiles

vigejolla:


vigejolla:
Buteo comes with it's own documentation, but that is not published anywhere. The idea here is not to duplicate the existing documentation in it's entirety, but give a short description of what a profile is and a place to link to from other places, e.g. accounts framework documentation.

pvuorela (approved):

PR #299, opened 2023-03-22
Updating the Setting up a Sailfish Phone document

jpwalden:


jovirkku (approved):
This shortened doc is probably what was requested.
I have nothing to change.

vigejolla (changes requested):
I think also the removed information is useful, and we can even expand it in the future. I'd hate to lose it completely. Please create a separate document for those already in this PR. I've been trying to think of a good name for this second part of the document, the best I've come up with so far is "Setting up a Sailfish Phone, part 2". You can use that or come up with a better name.

vigejolla:
Superseded by https://github.com/sailfishos/docs.sailfishos.org/pull/301

PR #310, opened 2023-04-04
Update README.md

ftigges-comdev:
lspci -nn | grep USB | cut -d '\[' -f3 | cut -d '\]' -f1  fails with error:

cut: the delimiter must be a single character


jovirkku:
Don't we have the same problem on line 62 as well, i.e., "the delimiter must be a single character"?

ftigges-comdev:
> Don't we have the same problem on line 62 as well, i.e., "the delimiter must be a single character"?

Yeah sorry.

Next time I just open an issue :D

ftigges-comdev:
I not familiar with github.
I close this pull request, change line 62 request again.

PR #311, opened 2023-04-04
Updating Email document

jpwalden:
Updated the doc with email certificate part, added to the end of the document

### DIFF
--- a/Support/Help_Articles/Email/README.md
+++ b/Support/Help_Articles/Email/README.md
@@ -244,6 +244,30 @@ The recipients of your messages will see your messages arriving from the account
 
 There are also the options to indicate the importance of the outbound message and to request read receipts.
 
+
+# Adding a new email certificate
+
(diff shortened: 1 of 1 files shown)


jovirkku (approved):
Looks good to me.

Support/Help_Articles/Email/README.md
```diff
@@ -244,6 +244,30 @@ The recipients of your messages will see your messages arriving from the account
 
 There are also the options to indicate the importance of the outbound message and to request read receipts.
 
+
+## Adding a new email certificate
```
vigejolla, line 248: Now this goes in as a subchapter of "[Composing and sending email](https://docs.sailfishos.org/Support/Help_Articles/Email/#composing-and-sending-email)". I think it would be better to have it as a separate chapter, i.e: 
```suggestion
# Adding a new email certificate
```

As a separate note (not blocker for merging this PR), wouldn't it make sense to also describe what the email certificates are used for, and why the user would want to add a new one?

jpwalden: Good points -thanks! Fixed the chapter. Also added some explanations why email certificate is needed etc.

vigejolla (approved):

PR #312, opened 2023-04-04
Add path to settings. Add picture IDs. Clarify.

jovirkku:
In addition to adding the path of the settings, I have done the following:

* shortened the links pointing to other Docs (made them local)
* added IDs for the pictures making it possible to point to the pictures with local hyperlinks
* added picture links to the text
* checked that I have consistent terminology (e.g."touch display lock")
* made it clear that the fingerprint detector of Xperia phones is not on the display

### DIFF
--- a/Support/Help_Articles/Device_Lock_and_Security_Code/README.md
+++ b/Support/Help_Articles/Device_Lock_and_Security_Code/README.md
@@ -9,22 +9,22 @@ nav_order: 350
 
 # About the Device lock
 
-The **Device lock** has three main purposes:
+The **Device lock** of Sailfish OS has three main purposes:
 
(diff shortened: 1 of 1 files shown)


jpwalden (commented):
On line 149: it is said: "NOTE:_  As the use..."
I assume that the _ is not meant to be there?
Otherwise document looks ok to me.

jovirkku:
> On line 149: it is said: "NOTE:_ As the use..." I assume that the _ is not meant to be there? Otherwise document looks ok to me.

The underscore is there to make the sentence appear in the italics font. For an unknown reason, one extra space character had destroyed the setup. Fixed now.

vigejolla (approved):

PR #316, opened 2023-04-06
Updating Bluetooth Pairing document

jpwalden:
Added supported BT profiles in the end of the document. Decided to add link to the document  which already exists: https://docs.sailfishos.org/Reference/Core_Areas_and_APIs/Networking/#bluetooth
That document however could be at some point be updated so that some of the missing profiles will be added there.

### DIFF
--- a/Support/Help_Articles/Bluetooth_Pairing/README.md
+++ b/Support/Help_Articles/Bluetooth_Pairing/README.md
@@ -215,4 +215,11 @@ devel-su
 
 The paired device appears in the Bluetooth menu of your phone now.
 
+# Bluetooth profiles supported by Sailfish OS
+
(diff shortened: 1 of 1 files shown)


Support/Help_Articles/Bluetooth_Pairing/README.md
```diff
@@ -215,4 +215,13 @@ devel-su
 
 The paired device appears in the Bluetooth menu of your phone now.
 
+# Bluetooth profiles supported by Sailfish OS
+
+[BlueZ](http://www.bluez.org/) is the base library for all Bluetooth-related features.
```
vigejolla, line 220: This is already mentioned in the document linked below, so isn't it kind of redundant to mention it here?

vigejolla (approved):

Support/Help_Articles/Bluetooth_Pairing/README.md
jpwalden: True, removed :)

PR #317, opened 2023-04-11
Updating Networking document

jpwalden:
Missing BT profiles added. (Those profiles added which were agreed with Jusa to be ok to be added)


jovirkku (approved):
Looks ok and correct to me.

vigejolla (approved):

PR #318, opened 2023-04-11
fix CLI commands in the Notes doc

jovirkku:


jpwalden (approved):
Changes looks good to me, now the commands are clearer.

vigejolla (approved):

PR #319, opened 2023-04-11
Fix nav orders l to m jb60521

jovirkku:
This PR is primarily about rearranging the navorders of docs beginning with L and M letters.

The two "Moving files..." docs had some links to fix, too. They had a bit different naming policy which bothered me and affected the alphabetical order


jpwalden (approved):
Looks good to me

vigejolla (approved):

PR #321, opened 2023-04-14
Document unsupported WEP encryption

nephros:
References:

 - https://forum.sailfishos.org/t/problem-with-wep-in-the-wlan-configuration-2-version/15237/15
 - https://github.com/sailfishos/wpa_supplicant/blob/b5123e537a51b136bcfa087c77d2ec0d90f5d004/rpm/build-config#L625


vigejolla (approved):
Thanks. FYI, I also filed a bug in our internal tracker about removing WEP from the UI.

PR #325, opened 2023-04-18
Add checklist for Sailfish installation

jovirkku:
This checklist is written to help those struggling with the bootloader, fastboot tools and mode, making the installation directory ready and flashing. 

It should help if something has gone wrong, and it could provide some more details than the official Sailfish X instructions.

On the other hand, this document has grown quite long although I tried to squeeze it. Let's hope that each situation's essential actions and options are covered.

One substantial improvement (I hope) was that I added tags to the chapters of various Sailfish X instruction documents (there are 15 of them for Xperia phones) so that we can give direct pointers from this checklist to the appropriate chapters of those instructions (e.g. for Xperia 10 II chapter 8 of the Windows instruction). This made this Checklist more clickable - previously the style was "open this document and go to chapter number 8".


jovirkku:
More convenient to read here: https://github.com/jovirkku/docs.sailfishos.org/tree/checklistSailfishOS-jb60552/Support/Help_Articles/Checklist_for_Sailfish_OS

Perhaps...

jpwalden (approved):
Very good document and also very useful one!
Only thing which might be worth of adding is that fastboot mode is explained/referred couple of times, but at least I didn't spot that there would be an explanation how the device can be put on fastboot mode. Naturally this is explained in the flashing instructions, however I think that could be added here as well, for example in the Introduction chapter. That is not mandatory, but at least if some other updates will be done that could be added as well.

vigejolla (approved):

jovirkku:
I have instructed how to enable the fastboot mode in the Windows section and at the 1st step of Troubleshooting (at the end of the doc). There are pointers to the latter from other parts of the document.

PR #329, opened 2023-04-20
Updating the Calendar document

jpwalden:
Updating the Calendar document with Deleting calendar database forcefully -article. Added to the end of the doc.

### DIFF
--- a/Support/Help_Articles/Calendar/README.md
+++ b/Support/Help_Articles/Calendar/README.md
@@ -178,4 +178,28 @@ Trigger a sync cycle by opening up "Settings > Accounts". Then long-tap the acco
 ## Collecting of logs
 This [article](https://jolla.zendesk.com/hc/en-us/articles/115011963227) explains how to collect logs concerning the sync of various accounts.
 
+## Deleting calendar database forcefully
+
+This article may help in deleting ghost calendars. With a ghost calendar we mean a calendar that persists in the Calendar app, no matter what you do to remove it.
+Unfortunately it will delete all valid calendars too, however deleting of the valid calendars is not critical if the those are synced calendars, since the calendar data will be synced from the server after the deletion. Only the data of the local (non-synced) calendars will be lost.
+
+### Enable developer mode
+
+You will need the Developer Mode and the password for root access & SSH.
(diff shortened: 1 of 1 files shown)


jovirkku (commented):
I wonder if line 183 would need some more information.

With a ghost calendar we mean a calendar that persists in the Calendar app, no matter what you do to remove it.

Maybe tell them that deleting the valid calendars is not critical if they are synced calendars as the calendar data would be synced down from the server(s) after this treatment. Only the data of the local (non-synced) calendars will be definitely lost.

Otherwise, this is "LGTM".

jpwalden:
> I wonder if line 183 would need some more information.
> 
> With a ghost calendar we mean a calendar that persists in the Calendar app, no matter what you do to remove it.
> 
> Maybe tell them that deleting the valid calendars is not critical if they are synced calendars as the calendar data would be synced down from the server(s) after this treatment. Only the data of the local (non-synced) calendars will be definitely lost.
> 
> Otherwise, this is "LGTM".

Good points, added those additional info to the doc.

Support/Help_Articles/Calendar/README.md
```diff
@@ -178,4 +178,28 @@ Trigger a sync cycle by opening up "Settings > Accounts". Then long-tap the acco
 ## Collecting of logs
 This [article](https://jolla.zendesk.com/hc/en-us/articles/115011963227) explains how to collect logs concerning the sync of various accounts.
 
+## Deleting calendar database forcefully
+
+This article may help in deleting ghost calendars. With a ghost calendar we mean a calendar that persists in the Calendar app, no matter what you do to remove it.
+Unfortunately it will delete all valid calendars too, however deleting of the valid calendars is not critical if the those are synced calendars, since the calendar data will be synced from the serverafter the deletion. Only the data of the local (non-synced) calendars will be lost.
```
vigejolla, line 184:
```suggestion
Unfortunately it will delete all valid calendars too, however deleting of the valid calendars is not critical if the those are synced calendars, since the calendar data will be synced from the server after the deletion. Only the data of the local (non-synced) calendars will be lost.
```

jpwalden: fixed

vigejolla (approved):

PR #332, opened 2023-04-24
Create Collect Camera Logs document

jpwalden:
Created collect camera logs document. Tested the script and works ok.

### DIFF
--- a/Support/Help_Articles/Collecting_Logs/Collect_Camera_Logs/README.md
+++ b/Support/Help_Articles/Collecting_Logs/Collect_Camera_Logs/README.md
@@ -1,0 +1,20 @@
+---
+title: Collect Camera Logs
+permalink: Support/Help_Articles/Collecting_Logs/Collect_Camera_Logs/
+grand_parent: Help Articles
+parent: Collecting Logs
+layout: default
+nav_order: 650
+---
+
+# Collecting camera logs
+
+In case you run into a problem with camera, run the command below at the Terminal as a normal user. After that use the camera application and reproduce the problem with the camera. Once you have done that you can study the log file or send it for analysis.
+
+```
+GST_DEBUG=4 jolla-camera > camera-log.txt 2>&1
+```
+
+The logs are saved into file camera-log.txt.
+
+


jovirkku (approved):
This is ok.
However, maybe it is unclear if the given command collects the logs right away, so from the past testing rounds, or if it enables the debugging mode meaning that a new test round should be run.

I think it is the latter, so a debugging mode is enabled.

I would propose that you write a couple of sentences telling that the command enabled the debugging and the user should now try to reproduce the issues with the camera. Once done, collect the log file and send it for analysis.

This may be plain cear for many users but perhaps not to all. I donno...

jpwalden:
> This is ok. However, maybe it is unclear if the given command collects the logs right away, so from the past testing rounds, or if it enables the debugging mode meaning that a new test round should be run.
> 
> I think it is the latter, so a debugging mode is enabled.
> 
> I would propose that you write a couple of sentences telling that the command enabled the debugging and the user should now try to reproduce the issues with the camera. Once done, collect the log file and send it for analysis.
> 
> This may be plain cear for many users but perhaps not to all. I donno...

Good point, since it actually works just as you expected. Added little bit more info how to use the script.

Support/Help_Articles/Collecting_Logs/Collect_Camera_Logs/README.md
```diff
@@ -0,0 +1,21 @@
...
+grand_parent: Help Articles
+parent: Collecting Logs
+layout: default
+nav_order: 650
+---
+
+# Collecting camera logs
+
+Collect logs from the Camera app log by running the following command at the Terminal as a normal user.
+In case you run into a problem with camera run the command below and then use the camera application and reproduce the problem with the camera. Once you have done that you can study the log file or send it for analysis.
```
vigejolla, line 13: I would combine these two lines, something like this:

In case you run into a problem with camera, run the command below at the Terminal as a normal use. After that use the camera application and reproduce the problem with the camera. Once you have done that...

jpwalden: Thanks, updated accordingly.

jovirkku: LGTM

vigejolla: You slightly misunderstood me - now, after the changes, the line 12 is redundant and could be removed completely.

jpwalden: Ok, yep the line 12 is not really needed -> removed.

vigejolla (approved):
LGTM

PR #334, opened 2023-04-25
Add guidance how to export contacts to VCF

jovirkku:


Support/Help_Articles/People_App_and_Contacts/README.md
```diff
@@ -276,4 +275,50 @@ The settings allow for changing the sorting order and if the last or first name
 
 The Phone app has three subpages. One of them is "People". It basically replicates the content and view of the People app, making it easier to choose people to call.
 
+# Exporting contacts to vCard file using the Terminal
+
+_Working on the command line safely requires that you have requied skills. Do it only if you feel confident about it._
```
vigejolla, line 280:
```suggestion
_Working on the command line safely requires that you have required skills. Do it only if you feel confident about it._
```

Support/Help_Articles/People_App_and_Contacts/README.md
```diff
@@ -276,4 +275,50 @@ The settings allow for changing the sorting order and if the last or first name
...
+Open the Terminal app and install the vcardconverter tool for the export:
+
+```
+devel-su
+pkcon refresh
+pkcon install nemo-qml-plugin-contacts-qt5-tools
+exit
+```
+
+The contacts are divided into collections in the phone database. Some examples of the collections are Local, SIM card and Aggregate. The first one contains the non-synced contacts, the 2nd those copied from the SIM card to the phone and the last one is a combination. You will need to choose which collection to export (if not all of them). There is unfortunately no "export all" command. Also, it may not be easy to guess by a collection name what it contains.
```
vigejolla, line 293: Doesn't the "aggregate" contain all contacts? I mean, isn't exporting it basically the same as "export all"?

Support/Help_Articles/People_App_and_Contacts/README.md
```diff
@@ -276,4 +275,50 @@ The settings allow for changing the sorting order and if the last or first name
...
+  ID: qtcontacts:org.nemomobile.contacts.sqlite::636f6c2d3131  Name: Contacts
+  ID: qtcontacts:org.nemomobile.contacts.sqlite::636f6c2d3132  Name: Suggested Contacts
+  ID: qtcontacts:org.nemomobile.contacts.sqlite::636f6c2d3133  Name: telepathy
+  ID: qtcontacts:org.nemomobile.contacts.sqlite::636f6c2d3134  Name: My Contacts
+```
+
+The actual export command is shown below followed by an example under it. This example shows how to export the aggregate collection (the 1st item in the list above).
+
+```
+devel-su -p vcardconverter --export <output-file> <address>
```
vigejolla, line 317:
```suggestion
devel-su -p vcardconverter --export <output-file> <ID>
```

vigejolla (changes requested):
some minor tweaks

Support/Help_Articles/People_App_and_Contacts/README.md
jovirkku: It looks like that on my phone, at least. Need to check on another phone. 
"Aggregate = all" would be good.

Support/Help_Articles/People_App_and_Contacts/README.md
jovirkku: Yes, ID is definitely better as it is used in the collection listing.

jovirkku: It turned out that the public releases contain the package "nemo-qml-plugin-contacts-qt5-tools". Therefore, I removed the chapter instructing how to install it.

vigejolla (approved):
LGTM

PR #336, opened 2023-04-26
Redirect from /Releases to /Support/Releases

vigejolla:


jovirkku (commented):
This looks technically correct. This won't add anything (unnecessarily) visible to the Documents view but the GitHub tree will have a fake Releases folder, right?

As we probably do not have any false/outdated links pointing to dos.sailfishos.org/Releases (other than those in the browser cache, possibly), how relevant a change is this?

rainemak (approved):

PR #337, opened 2023-04-26
Create Collect Location Services Logs document

jpwalden:
Location services logging document created, tested the described flow and worked ok.

### DIFF
--- a/Support/Help_Articles/Collecting_Logs/Collect_Location_Services_Logs/README.md
+++ b/Support/Help_Articles/Collecting_Logs/Collect_Location_Services_Logs/README.md
@@ -1,0 +1,48 @@
+---
+title: Collect Location Services Logs
+permalink: Support/Help_Articles/Collecting_Logs/Collect_Location_Services_Logs/
+grand_parent: Help Articles
+parent: Collecting Logs
+layout: default
+nav_order: 950
+---
+
+# Collect Location (geoclue) Services (GNSS) Logs
+
+GNSS = Global Navigation Satellite System
+
+Commonly used satellites: GPS, Glonass, Galileo, Beidou.
+
+In order to enable the location services logs follow these steps:
+1. Enable the Developer tools as instructed in this [help article](/Support/Help_Articles/Enabling_Developer_Mode/).
+2. Open the Terminal app on the phone (or use an SSH connection to the phone). 
+3. In order to get the super-user rights give this command:
+```
+devel-su
+```
+4. To enable the GNSS logging, Add /var/lib/environment/nemo/qtlogging.conf with the following command:
+```
+vi /var/lib/environment/nemo/qtlogging.conf
+```
+After that add this line to the file:
+```
+QT_LOGGING_RULES="geoclue.provider.*=true"
+```
+5. Reboot your device (or restart the user session).
+6. Open some application which is using location, for example GPSinfo or some Android application.
+7. Attach the journal log with these commands:
+```
+devel-su
+journalctl -a -b > journal-geoclue.txt
+```
+After these steps you have an journal log which includes location related logs. This can be verified by opening the journal log and checking that there are lines with the phrase "geoclue-hybris". This can be easily done with the command:
+```
+grep geoclue-hybris journal-geoclue.txt
+```
+
+8. After logging has been done you can disable the GNSS logging in order to avoid spamming the journal:
+```
+devel-su
+rm /var/lib/environment/nemo/qtlogging.conf
+reboot
+```


Support/Help_Articles/Collecting_Logs/Collect_Location_Services_Logs/README.md
```diff
@@ -0,0 +1,33 @@
...
+Commonly used satellites: GPS, Glonass, Galileo, Beidou.
+
+In order to enable the location services logs follow these steps:
+1. Enable the Developer tools as instructed in this [help article](/Support/Help_Articles/Enabling_Developer_Mode/).
+2. Open the Terminal app on the phone (or use an SSH connection to the phone). 
+3. In order to get the super-user rights give this command:
+```
+devel-su
+```
+4. To enable the GNSS logging, create the file /var/lib/environment/nemo/qtlogging.conf with the following command:
```
jovirkku, line 22: "Create the file....with the following command".
The command does not create any file. Instead, that command must be included in the conf file. The phrase looks unclear to me. I would change the text (once written by me?) to something like this:
"To enable the GNSS logging, create file /var/lib/environment/nemo/qtlogging.conf and include the following command in that file".

Support/Help_Articles/Collecting_Logs/Collect_Location_Services_Logs/README.md
```diff
@@ -0,0 +1,33 @@
...
+QT_LOGGING_RULES="geoclue.provider.*=true"
+```
+5. Reboot your device (or restart the user session).
+6. Open some application which is using location, for example GPSinfo or some Android application.
+7. Attach the journal log with these commands:
+```
+devel-su
+journalctl -a -b > journal-geoclue.txt
+```
+After these steps you have an journal log which includes location related logs. This can be verified by opening the journal log and checking that there are lines with the phrase "geoclue-hybris".
```
jovirkku, line 33: We could add the following sentence if you see it feasible (not mandatory, though)

This can be easily done with the command:
cat journal-geoclue.txt | grep geoclue-hybris

vigejolla: > We could add the following sentence if you see it feasible (not mandatory, though)
> 
> This can be easily done with the command: cat journal-geoclue.txt | grep geoclue-hybris

nitpick: Useless use of cat. `grep geoclue-hybris journal-geoclue.txt` works just as well.

jovirkku: I like cats!

vigejolla: Me too, that's why I don't want them to be used for unnecessary work :)

Support/Help_Articles/Collecting_Logs/Collect_Location_Services_Logs/README.md
jpwalden: Modified that a bit: "To enable the GNSS logging, modify /var/lib/environment/nemo/qtlogging.conf with the following command:
: QT_LOGGING_RULES="geoclue.provider.*=true"
I guess that is ok?

Support/Help_Articles/Collecting_Logs/Collect_Location_Services_Logs/README.md
jpwalden: Thanks for the tips, added some of those, however I assume that also this would be ok: QT_LOGGING_RULES="geoclue.provider.*=false ?

jovirkku: Have you tested that "false" works? Why should it be false and not true? "true" I have tested and it works.
I have no idea how the logging tool uses that variable.

jpwalden: I'll modify the instructions as discussed, I'll send a new update soonish

Support/Help_Articles/Collecting_Logs/Collect_Location_Services_Logs/README.md
```diff
@@ -0,0 +1,43 @@
...
+grand_parent: Help Articles
+parent: Collecting Logs
+layout: default
+nav_order: 950
+---
+
+# Collect Location (geoclue) Services (GNSS) Logs
+
+GNSS = Global Navigation Satellite System
+Commonly used satellites: GPS, Glonass, Galileo, Beidou.
```
vigejolla, line 13: On the rendered page there is no line break betweek "System" and "Commonly", so this looks a bit funny

Support/Help_Articles/Collecting_Logs/Collect_Location_Services_Logs/README.md
```diff
@@ -0,0 +1,43 @@
...
+After these steps you have an journal log which includes location related logs. This can be verified by opening the journal log and checking that there are lines with the phrase "geoclue-hybris". This can be easily done with the command:
+```
+grep geoclue-hybris journal-geoclue.txt
+```
+
+8. After logging has been done you can disable the GNSS logging in order to avoid spamming the journal:
+```
+devel-su
+QT_LOGGING_RULES="geoclue.provider.*=false"
+reboot
```
vigejolla, line 42: Now you are only setting the environment variable for the current session, so it does not have any effect after reboot.

Best would be to remove or comment out the line from qtlogging.conf.

vigejolla (changes requested):

Support/Help_Articles/Collecting_Logs/Collect_Location_Services_Logs/README.md
jpwalden: True, fixed.

Support/Help_Articles/Collecting_Logs/Collect_Location_Services_Logs/README.md
jpwalden: True, edited so that the qtlogging.conf will be removed since it is no longer needed.

Support/Help_Articles/Collecting_Logs/Collect_Location_Services_Logs/README.md
jpwalden: Edited so that it is clearer that the qtlogging.conf also needs to be added and modified.

vigejolla (approved):
LGTM

PR #339, opened 2023-04-28
Create Collect Driver and Display Logs document

jpwalden:
Collect Driver and Display Logs document created. Tested the commands etc. and added some, seems to work ok.

### DIFF
--- a/Support/Help_Articles/Collecting_Logs/Collect_Kernel_Logs/README.md
+++ b/Support/Help_Articles/Collecting_Logs/Collect_Kernel_Logs/README.md
@@ -1,0 +1,41 @@
+---
+title: Collect Kernel Logs
+permalink: Support/Help_Articles/Collecting_Logs/Collect_Kernel_Logs/
+grand_parent: Help Articles
+parent: Collecting Logs
+layout: default
+nav_order: 900
+---
+
+# Collecting and monitoring kernel logs
+With journalctl command it is possible to print out the message buffer of the [kernel](https://en.wikipedia.org/wiki/Kernel_(computer_science)). The output of this command typically contains the messages produced by the [device drivers](https://en.wikipedia.org/wiki/Device_drivers).
+
+First you need to enable the developer mode, instructions for that can be found from this [article](/Support/Help_Articles/Enabling_Developer_Mode/).
+Once that is done, then open the Terminal app of a Sailfish device or make an SSH connection to the device. Instructions about SSH can be found from this [article](/Support/Help_Articles/SSH_and_SCP/). Get the super-user rights (root) with this command:
+```
+devel-su
+```
+Start monitoring the logs with this command:
+```
+journalctl -k -f
+```
+
+The following command picks up fatal messages only:
+```
+journalctl -k -f | grep Fatal
+```
+
+It may be good to direct the flow to a file for later investigation, that can be done with this command:
+```
+journalctl -k -f > kernel.log
+```
+or with this:
+```
+journalctl -k -f | grep Fatal > kernel.log
+```
+
+Messages keep flowing to the screen or to the file until ```<ctrl>C``` is used to stop it.
+After you have stopped the message flow and if you have been directing the messages to log file, then you can check the content for example with this command:
+```
+cat kernel.log
+```


Support/Help_Articles/Collecting_Logs/Collect_Driver_and_Display_Logs/README.md
```diff
@@ -0,0 +1,42 @@
...
+title: Collect Driver and Display Logs
+permalink: Support/Help_Articles/Collecting_Logs/Collect_Driver_and_Display_Logs/
+grand_parent: Help Articles
+parent: Collecting Logs
+layout: default
+nav_order: 700
+---
+
+# Collecting and monitoring driver and display messages (dmesg)
+**dmesg** (display message or driver message) is a command on most [Unix-like](https://en.wikipedia.org/wiki/Unix-like) operating systems that prints the message buffer of the [kernel](https://en.wikipedia.org/wiki/Kernel_(computer_science)). The output of this command typically contains the messages produced by the [device drivers](https://en.wikipedia.org/wiki/Device_drivers).
```
vigejolla, line 11: The original [unix man page](http://man.cat-v.org/unix_8th/8/dmesg) calls dmesg "diagnostic messages".

vigejolla (commented):
There are some aspects about this article that I don't like.

First, the title. "Driver and Display Logs". The article is about kernel logs. Talking about "Driver and Display Logs" is confusing to anyone with any linux background. Sure, driver logs are there as well, as they are kernel modules, but what does "display" have to do with these?

Then, the approach. Why are we teaching `dmesg` command? Wouldn't it be better to teach `journalctl -k` ? The contents is the same, but dmesg only prints the contents of the kernel ring buffer. As the name implies, the buffer contents get overwritten after a while. systemd on the other hand contains logs for a prolonged period - especially when combined with [Persistent logs] (https://docs.sailfishos.org/Support/Help_Articles/Collecting_Logs/Collect_Persistent_Logs/).

Sure, dmesg was handy in the 90s, but we are already in 2023 ;)

Support/Help_Articles/Collecting_Logs/Collect_Driver_and_Display_Logs/README.md
jpwalden: Thanks for the comments, should we rename this then to "collect kernel logs"?
We could also replace the dmesg command with journalctl -f ?

vigejolla: "Collect kernel logs" sounds good to me. And yes, I would replace with `journalctl -k -f`.

jpwalden: Updated

Support/Help_Articles/Collecting_Logs/Collect_Kernel_Logs/README.md
```diff
@@ -0,0 +1,41 @@
...
+parent: Collecting Logs
+layout: default
+nav_order: 900
+---
+
+# Collecting and monitoring kernel logs
+With journalctl command it is possible to print out the message buffer of the [kernel](https://en.wikipedia.org/wiki/Kernel_(computer_science)). The output of this command typically contains the messages produced by the [device drivers](https://en.wikipedia.org/wiki/Device_drivers).
+
+First you need to enable the developer mode, instructions for that can be found from this [article](/Support/Help_Articles/Enabling_Developer_Mode/).
+Once that is done, then open the Terminal app of a Sailfish device or make an SSH connection to the device. Get the super-user rights (root] with this command:
```
jovirkku, line 14: a typo to fix:  (root]
If we want to add a link to SSH instructions, this would be it: /Support/Help_Articles/SSH_and_SCP/

jovirkku (approved):

Support/Help_Articles/Collecting_Logs/Collect_Kernel_Logs/README.md
```diff
@@ -0,0 +1,41 @@
...
+It may be good to direct the flow to a file for later investigation, that can be done with this command:
+```
+journalctl -k -f > kernel.log
+```
+or with this:
+```
+journalctl -k -f | grep Fatal > kernel.log
+```
+
+Messages keep flowing to the screen until <ctrl>C is used to stop it.
```
vigejolla, line 37: Nitpicking: The messages will not flow to the screen because we have just directed the flow to a file.

vigejolla (commented):
Almost ready to merge: Please fix the (root] typo that jovirkku mentioned, and while at it, you might consider the wording in the <ctrl-c> paragraph.

Support/Help_Articles/Collecting_Logs/Collect_Kernel_Logs/README.md
jpwalden: Good catch, fixed. Added also the link.

Support/Help_Articles/Collecting_Logs/Collect_Kernel_Logs/README.md
jpwalden: Valid point, added related wording since that part is about all of the above commands.

PR #344, opened 2023-05-08
Instruct how to add a new folder to media indexing

jovirkku:
Let's see.

There was a discussion on whether this change is necessary as it is volatile in terms of permissions. Namely, a Sailfish OS update may wipe the added permissions away from /etc/sailjail/permissions.

The text has now a warning on this.


vigejolla (approved):
LGTM

PR #345, opened 2023-05-09
Deleting BT devices using the command line

jovirkku:
To be honest, I have no idea how often the added procedure would be needed.
Take it or leave it.

There is something to clean up in the other chapters of this document but let it be another commit.


jpwalden (approved):
Checking and testing, LGTM.

vigejolla (approved):
LGTM

PR #351, opened 2023-05-12
Updating Storage Space document

jpwalden:
Added How to check the big disc space consumers -documents content to Storage Space -document.

### DIFF
--- a/Support/Help_Articles/Storage_Space/README.md
+++ b/Support/Help_Articles/Storage_Space/README.md
@@ -49,6 +49,40 @@ devel-su
 du -xm /home | sort | tail -20 |  sort -nr
 ```
 
+## How to check the big disc space consumers
+
+The following commands may be helpful in checking what might be wrong with the file system. This information may help in resolving a problem in getting Sailfish OS updates.
+
+In many cases, the problem is found in some files that the user has installed to the system data partition, which then prevents installation of the update. This is usually because of some hacking/development which has brought stale files to the system data partition or the installation of packages that consume the disc space from the system data partition.
+Each case looks a bit different but was caused by the user doing things in the developer mode (without understanding the consequences).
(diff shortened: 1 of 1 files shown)


Support/Help_Articles/Storage_Space/README.md
```diff
@@ -49,6 +49,39 @@ devel-su
...
+```
+du -a -x / | awk '{if($1 > 102400) print int($1/1024) "MB" " " $2 }' | sort -nr
+```
+```
+du -a -x /home | awk '{if($1 > 102400) print int($1/1024) "MB" " " $2 }' | sort -nr
+```
+
+This command presents the installed code (rpm) packages, listed by size:
+```
+rpm -qa --queryformat '%{SIZE}\t %{NAME}-%{version}\n' | sort -nr | more
```
vigejolla, line 82: I kind of have a feeling that we should have some kind of warning here, along the lines of "Please note that removing packages can break your system if you don't know what you are doing". Or something.

vigejolla (changes requested):

Support/Help_Articles/Storage_Space/README.md
```diff
@@ -49,6 +49,39 @@ devel-su
...
+
+devel-su
+```
+
+A rough listing (in reverse order) to begin with. It gives you an idea of where the storage space consumers are, by folder size.
+```
+du -xa / | sort -n
+```
+
+A more fine-grained listing, both from the root and from home:
```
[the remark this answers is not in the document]
jpwalden: Yes, you are right. Changed the order since makes sense to check the high-level list first and after that if needed the full list.

Support/Help_Articles/Storage_Space/README.md
jpwalden: Valid point, added a warning.

vigejolla (approved):

PR #359, opened 2023-05-24
Updating Whatsapp Setup document

jpwalden:
Updated how to move WhatsApp backup to another phone part of the document


Support/Help_Articles/Whatsapp_Setup/README.md
```diff
@@ -140,68 +140,73 @@ The settings button makes it possible to tune app the Android-specific settings
...
-________
+ls -l /home/.android/data/data | grep com.android.providers.media
+
+This command prints a line like that below:
+drwx------  6 510006            510006 4096 Mar 14 17:32 com.android.providers.media
 
-(*) We are looking for alternative ways to do this easily.
+On that line, the number 510006 is your UID. We will need it in the next command. Use the numric code in the command below.
+
+chown -R UID:media_rw android_storage/Android/media/com.whatsapp/WhatsApp
```
vigejolla, line 208: Two things:
1) I wonder if it's clear enough that "UID" is supposed to be replaced with the number. I would use `chown -R <UID>:media_rw android_storage/Android/media/com.whatsapp/WhatsApp` here, and on the line above I would explicitly state that "<UID>" is supposed to be replaced.
2) Is  "media_rw" really correct? On my device the group of these files seem to be "appsupport-media_rw", which sounds more correct to me. Also the owner on my device is not 51006 (which matches dir com.android.providers.media in /home/.android/data/data) but 510087 (which matches dir com.android.providers.media.module).

Support/Help_Articles/Whatsapp_Setup/README.md
```diff
@@ -140,68 +140,73 @@ The settings button makes it possible to tune app the Android-specific settings
...
 
-Done.
-________
+ls -l /home/.android/data/data | grep com.android.providers.media
+
+This command prints a line like that below:
+drwx------  6 510006            510006 4096 Mar 14 17:32 com.android.providers.media
 
-(*) We are looking for alternative ways to do this easily.
+On that line, the number 510006 is your UID. We will need it in the next command. Use the numric code in the command below.
```
vigejolla, line 206: typo in numeric

vigejolla (changes requested):
At least fix the typo. And please double check the UID:GID of the chown command.

Support/Help_Articles/Whatsapp_Setup/README.md
jpwalden: 1. Valid point, updated.
2. Good comment, however I'm not fully familiar with this. On my testing this part wasn't needed so I couldn't verify that how these commands really affects. Anyways updated this part as you proposed.

vigejolla (approved):
I *believe* this is now in working state. I'm not 100% sure though. I'll merge this as it is now, let's fix later if we find issues.

PR #361, opened 2023-05-30
Crosslinking diagnostic sections

thigg:
I just looked for the diagnostic commands and noticed there are multiple places where diagnostic commands are listed. As a quickfix i propose to link those to each other. The android support section should do this as well probably

### DIFF
--- a/Reference/Sailfish_OS_Cheat_Sheet/README.md
+++ b/Reference/Sailfish_OS_Cheat_Sheet/README.md
@@ -141,6 +141,10 @@ zypper remove --clean-deps PACKAGENAME
 
 ## Diagnostics
 
+
+
+Also see [this section](https://docs.sailfishos.org/Develop/Platform/Testing_Advic)
+
 Saving logs is always good
 ```nosh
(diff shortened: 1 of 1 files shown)


Reference/Sailfish_OS_Cheat_Sheet/README.md
```diff
@@ -141,6 +141,10 @@ zypper remove --clean-deps PACKAGENAME
 
 ## Diagnostics
 
+
+
+Also see [this section](https://docs.sailfishos.org/Develop/Platform/Testing_Advic)
```
vigejolla, line 146: Generally speaking, I think adding cross reference is a good idea. However, 1) There is typo in the link (missing 'e' in the end) and 2) It's better to not have the protocol and host sections in internal links, so that the links work also when someone is running the site locally (like I do when reviewing PRs). So:

```suggestion
Also see [this section](/Develop/Platform/Testing_Advice)
```

vigejolla (approved):
LGTM

vigejolla:
Thank you!

PR #363, opened 2023-05-31
Use full path to kill camera provider

rainemak:
Logs from terminal:
> killall android.hardware.camera.provider@2.4-service_64
> android.hardware.camera.provider@2.4-service_64: no process found

Instead, let's use full path.


vigejolla (approved):
LGTM

PR #365, opened 2023-05-31
Update README.md

poetaster:
using python3-imaging will get into harbour. using pkconfig(pillow) will not.

I just tested this with a new release of Imageworks.


vigejolla:
> using python3-imaging will get into harbour. using pkconfig(pillow) will not.
> 
> I just tested this with a new release of Imageworks.

Tested how? The Imageworks package which is still in the harbour queue gives me:
```
ERROR [python3-imaging] Dependency not allowed
```
as it should. Use python3dist(pillow) instead.

btw: this page is automatically generated, so it matches exactly the current harbour requirements.

vigejolla:
> using python3-imaging will get into harbour. using pkconfig(pillow) will not.

Now I noticed your error: It's `python3dist(pillow)`, not `pkgconfig(pillow)`.

PR #367, opened 2023-06-07
Small tweaks

rainemak:
This PR adds create-bundle (maybe should be called run-bundle) that combines commands that are document in [README.md](https://github.com/sailfishos/docs.sailfishos.org/blob/master/README.md#preview-on-your-desktop). In addition, this  commit adds 'vendor/bundle' to git ignore as per script & [README.md](https://github.com/sailfishos/docs.sailfishos.org/blob/master/README.md#preview-on-your-desktop)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,3 +6,4 @@
 _site
 Gemfile.lock
 node_modules
+vendor/bundle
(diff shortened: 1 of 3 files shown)


vigejolla:
Should we change README.md as well, so that it would point to the new script instead of individual commands?

rainemak:
> Should we change README.md as well, so that it would point to the new script instead of individual commands?

Updated.

vigejolla (approved):
LGTM

PR #374, opened 2023-06-28
Add document about signing packages

vigejolla:


Develop/Apps/Packaging/Signing_Packages/README.md
```diff
@@ -0,0 +1,66 @@
...
+---
+
+# Signing RPM packages
+
+## Creating a key for signing
+
+RPM packages are signed using GPG keys. In order to be able to sign RPM packages, a suitable key needs to be created. This can be done using standard `gpg` tool:
+
+```
+$ gpg --full-generate-key
```
Thaodan, line 17: gpg --gen-key should be enough, no need to use the option to asks for ALL options.

Develop/Apps/Packaging/Signing_Packages/README.md
```diff
@@ -0,0 +1,66 @@
...
+
+The tool will ask all the necessary information. When unsure, select the default value.
+
+### Passphrase file
+
+If your gpg key is protected with a passphrase, you need to provide it to the sfdk tool via passphrase file. This is just a text file containing the passphrase. The permissions of the file should be 600, so that it’s not readable by anyone else.
+
+```
+$ echo “This is my passphrase” > $HOME/passphrase.txt
+$ chmod 0600 $HOME/passphrase.txt
```
Thaodan, line 28: I would add a note here that this is just an example path, the user should choose their on path.

Reference/Software_Packaging/Signing_Packages/README.md
```diff
@@ -0,0 +1,11 @@
...
+title: Signing Packages
+permalink: Reference/Software_Packaging/Signing_Packages
+parent: Software Packaging
+layout: default
+nav_order: 200
+---
+
+The Sailfish SDK can be used for signing RPM packages - we have a [separate page](/Develop/Apps/Packaging/Signing_Packages) describing the usage. 
+
+The package signatures are currently not used in Sailfish OS, but you can still use them for manually verifying the package contents.
```
Thaodan, line 11: This is not entirely true if the user instructs the user of their packages to import their public it can be used.

`rpm --import keyfile.gpg` has to be done on the target device.

Thaodan (changes requested):
Looks good mostly except a few small comments.

Reference/Software_Packaging/Signing_Packages/README.md
vigejolla: > This is not entirely true if the user instructs the user of their packages to import their public it can be used.
> 
> `rpm --import keyfile.gpg` has to be done on the target device.

Because of the bug that does not really help much. Well, it does help that manually verifying the signature is possible. I'd rather not add instructions about that yet.

vigejolla: Thinking again, I decided to add the link to the other page which describes how verification is done - the `rpm --import` command is mentioned there.

Thaodan: Makes sense.

Thaodan (approved):
LGTM.

PR #375, opened 2023-07-16
repair and simplify sync log collection page.

dryo:
I'm not sure if `tee` is included in a standard installation, though.

### DIFF
--- a/Support/Help_Articles/Collecting_Logs/Collect_Synchronization_Logs/README.md
+++ b/Support/Help_Articles/Collecting_Logs/Collect_Synchronization_Logs/README.md
@@ -63,7 +63,7 @@ devel-su -p
 QTCONTACTS_SQLITE_TWCSA_TRACE=1 QTCONTACTS_SQLITE_TRACE=1 MSYNCD_LOGGING_LEVEL=8 msyncd 2>&1 | tee msyncd.log
 ```
 3. Trigger a sync cycle by opening up "Settings > Accounts". Then long-press the account you want to debug. Tap Sync in the pop-up menu.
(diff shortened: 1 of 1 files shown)


dryo:
See https://forum.sailfishos.org/t/ist-the-documentation-for-getting-caldav-carddav-sync-logs-faulty/16220 why the `devel-su` needs a `-p`.

Thaodan (commented):
Why the changes in the name of the headings?

Thaodan (changes requested):
In general I agree with the changes but there is a better way to do this.

1. `systemctl --user edit msyncd`
2. Paste this into the opened editor:
```
[Service]
Environment=QTCONTACTS_SQLITE_TRACE=1 
Environment=MSYNCD_LOGGING_LEVEL=8 
Environment=QTCONTACTS_SQLITE_TWCSA_TRACE=1
```
3. systemctl --user restart msyncd 
4. journalctl --user --boot --unit msyncd.service 
5. If the --user filter doesn't work try /usr/bin/msyncd instead of --user ---unit msyncd.service

To revert remove `rm ~/.config/systemd/user/msyncd.service.d/override.conf`

Doing so will ensure that the logfile will only contain output from the msyncd service.

In general calling journalctl to log something from the user session without filters is bad as it  contains to much irrelevant information.

jovirkku:
See [this post](https://forum.sailfishos.org/t/ist-the-documentation-for-getting-caldav-carddav-sync-logs-faulty/16220/5) why to prefer 'tee' to 'cat'.

jovirkku (approved):
LGMT.

jovirkku (commented):
The proposed changes make sense and look good to me.

PR #379, opened 2023-09-20
Update links to Sony's flash-tool "Emma"

Olf0:
Because the old links have become broken by Sony.

Additionally soften the langauge, because Emma is not "required" to flash an Xperia phone, other flash tools exist, which also run under MacOS and Linux: XperiFirm, Flashtool by Androxyde, Newflasher.


Olf0:
Please not that the link for unlocking an Xperia's bootloader also was altered, it is now at https://developer.sony.com/open-source/aosp-on-xperia-open-devices/get-started/unlock-bootloader#jss-server-side

You might want to change this bootloader link (and maybe the one for "Emma", too) at other locations, e.g., https://jolla.com/sailfishxinstall/

mlehtima (approved):

PR #390, opened 2024-01-29
APKPure document updated

jovirkku:
The trigger for this update was that the URL of APKPure was changed from apkpure.com to apkpure.net.

While working, I noticed that the APKPure app itself had been changed too, and hence our screenshots and text lines were not exactly right in some places. Therefore, I updated a lot, while checking with 4.5.0 and 4.6.0 releases at the same time.
I have also numbered the pictures so that it would be easier to point to them.

I checked the text with the Grammarly tool to eliminate stupid language mistakes.

Note: Even if the URL extension of the website is now ".net", the downloaded APK files still have ".com" in their names.

Unfortunately, the website of APKpure is loaded with advertisements. They use the same colours and style as the actual page content. I did my best to guide the reader to the correct download button.  :-)

There are too many changes, sorry. It may be easier to see the outcome by looking at this:
https://github.com/jovirkku/docs.sailfishos.org/tree/fixApkPureURL-jb61429/Support/Help_Articles/Android_App_Support/APKPure 

Sadly, it does not show how it looks. There are no pictures side-by-side, for instance. Only Jekyll can do it.


jpwalden (approved):
Looks good to me, really comprehensive update with really good pictures and step-by-step instructions!

rainemak (approved):

PR #395, opened 2024-03-25
Remind on untrusted SW setting for AppSupport. Fix broken links.

jovirkku:
The documents on AppSupport in general, on APKPure and WhatsApp, did not mention the requirement of enabling "Allow untrusted software". This is fixed in this PR. In addition, the AppSupport article had some broken URLs.


rainemak (approved):
Add minor note but approved already.

PR #409, opened 2024-04-22
Small adjustments to i18n

pvuorela:
Remove reference to layout direction. We don't support now other than left-to-right languages, and right-to-left ones would likely anyway be laid out mostly the same. And if not the direction should be anyway coming from the script info, not translations.


jovirkku (approved):
Looks ok to me.

keto (approved):

PR #413, opened 2024-06-18
Add audio and keyboard reset to Utilities

jovirkku:
The Utilities of the OS release 4.6.0 has the following changes:
- audio reset added
- keyboard reset added
Two screenshots updated.


rainemak (approved):
LGTM

PR #417, opened 2024-08-26
Provide direct links to latest tooling and target images

martyone:
It is inconvenient to search for these in the directory listing behind the original link, which does not do the right sorting and truncates the long file names, making the distinguishing part visible only on mouse hover.


martyone:
This is still not ideal as it does not links the EA images directly and omits the actual release numbers, but I would prefer to not add more content that would have to be kept up-to-date manually.

jovirkku:
This is indeed a necessary and helpful improvement. LGTM.

jovirkku (approved):

PR #420, opened 2024-09-05
Suggest to install docker-buildx on Ubuntu

martyone:
We have seen how Docker installed through the docker.io package broke by upgrade, likely due to https://bugs.launchpad.net/ubuntu/+source/docker-buildx/+bug/1968035. Installing docker-buildx seems more appropriate.


jovirkku (approved):
LGTM. 
I could fix my Docker in this way, enabling a successful reinstallation of the SDK.

PR #423, opened 2024-09-29
Add documentation about OBS Packaging and building

nephros:
Co-authored-by: sebix

Initial Version, from the Prague Sailathon 2024


keto (commented):
Added some initial comments. (sorry for not looking into this earlier)

In general a really good start. There are few things that are maybe not specific to OBS and could go into documents about [packaging and packaging formats](https://docs.sailfishos.org/Tools/Sailfish_SDK/Building_packages/#packaging-formats) .

PR #434, opened 2025-01-30
Add BoF community meeting at FOSDEM

abranson:
Add FOSDEM BoF meeting


rainemak:
If I import this here in Finland, would the local time @ Brussels be right?

image

vs

image

abranson:
Yeah should be. You're an hour ahead of Brussels. Your calendar entries should all migrate when you land.

PR #458, opened 2026-03-16
Fix signing instructions

attah:
Remove silly-quotes and fix config syntax.

Also `package.signing-passphrase` should be mentioned, but it appears to be broken.


attah:
Hmm, this is weird. The original config commands should work according to the error message i get when trying - but they very much don't work.

rainemak (approved):

PR #459, opened 2026-04-10
Browser: add a section about certificates

nephros:
Add a section about importing TLS certificates into the local profile.

### DIFF
--- a/Reference/Core_Areas_and_APIs/Browser/Working_with_Browser/README.md
+++ b/Reference/Core_Areas_and_APIs/Browser/Working_with_Browser/README.md
@@ -6,6 +6,48 @@ grand_parent: Core Areas and APIs
 layout: default
 ---
 
+## The Browser Profile
+
+The Sailfish OS Browser stores its profile at `${HOME}/.local/share/org.sailfishos/browser/.mozilla/`
+
+Some of the operations for configuring this profile given in the official
+Mozilla Documentation are supported.
+
+Such as:
+
+### Adding TLS Certificates
+
+User-specific CA or Client TLS certificates can be added to the user-specific certificate store:
+
+For a CA Certificate:
+
+```
+    export CA_NAME="My Root Authority"
+    export PEM=/path/to/ca.pem
+    export MOZILLA_PROFILE="${HOME}/.local/share/org.sailfishos/browser/.mozilla/"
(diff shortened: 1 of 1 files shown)


Reference/Core_Areas_and_APIs/Browser/Working_with_Browser/README.md
```diff
@@ -6,6 +6,48 @@ grand_parent: Core Areas and APIs
...
+    export PEM=/path/to/client.pem
+    export MOZILLA_PROFILE="${HOME}/.local/share/org.sailfishos/browser/.mozilla/"
+    certutil -A -n "$CRT_NAME" -t "Pu,," -d "${MOZILLA_PROFILE} -i ${PEM}
+```
+
+Note the differences in the parameter to the `-t` option!
+
+The `certutil` utility is provided by the `nss-tools` package.
+
+**Note:** in Sailfish OS releases prior to version 4.x, the Browser profile was stored in ` -d ${HOME}/.mozilla/mozembed/`
```
pvuorela, line 49: Is this relevant anymore?

Reference/Core_Areas_and_APIs/Browser/Working_with_Browser/README.md
```diff
@@ -6,6 +6,48 @@ grand_parent: Core Areas and APIs
 layout: default
 ---
 
+## The Browser Profile
```
pvuorela, line 9: Is there reason to make these now first items in the page? Could think for browser development the logging and building type of things would still be more important.

nephros: I see what you mean. This is more a user guide vs. development documentation.  
Would you be open to reorganize the Browser-related pages into e.g.

 - Browser
   - Usage and Configuration
   - Performance Testing
   - Building
   - Debugging

So it is clearer what the pages are for?

pvuorela: Well at the moment, or even with these changes, the content here is hardly a user guide. If we wanted such for Browser, think that should go then somewhere around these https://docs.sailfishos.org/Support/Help_Articles/ - though these certificate things could be too low level there. Generally user guide should rather document what's available on the UI.

Fwiw, elsewhere this page is now referred as "browser development page" - https://docs.sailfishos.org/Reference/Core_Areas_and_APIs/Browser/

Is the motivation for adding these texts as app user (not developer) working around not having a direct UI for it?

PR #460, opened 2026-04-13
Mozilla location service is dead, replace references with beaconDb

pvuorela:
For now just removed telling explicitly how the advanced location settings looks like, and the old screenshot from it. The actual content might depend a bit on the device etc.


Reference/Core_Areas_and_APIs/Apps_and_MW/Positioning/README.md
```diff
@@ -25,7 +25,7 @@ There are some Sailfish OS-specific GeoClue provider plugins which provide posit
 The code related to the location and positioning stack can be found in the following repositories:
 
   - geoclue-provider-hybris, the main GeoClue provider plugin, using libhybris interfaces to control the GPS: <https://github.com/mer-hybris/geoclue-providers-hybris>
-  - geoclue-provider-mlsdb, a secondary GeoClue provider plugin, using Mozilla Location Service to provide assisted-GPS position updates: <https://github.com/mer-hybris/geoclue-providers-mlsdb>
+  - geoclue-provider-mlsdb, a secondary GeoClue provider plugin. Was earlier using now-defunct Mozilla Location Services to provide assisted-GPS position updates. Nowadays it's using API-compatible [beaconDB](https://beacondb.net/) service. Code available at: <https://github.com/mer-hybris/geoclue-providers-mlsdb>
```
direc85, line 28: I don't think we need to explain the relation to MLS here. At least not the first thing, maybe after the link, but I'd rather have it all in the BeaconDB chapter in the readme.

direc85 (approved):
One suggestion but it's not wrong by any means, so approving already.

Reference/Core_Areas_and_APIs/Apps_and_MW/Positioning/README.md
pvuorela: Since this is slightly more on the technical side, guess I was trying to explain the naming mismatch. Agree it's not totally mandatory, though not totally out of place either.

PR #465, opened 2026-05-23
Warn about truncated data in CSV calendar export

marhop:
Hi,

I noticed that since the recent sqlite v3.52 release it's possible for long values to be truncated when exporting the calendar to CSV. This PR adds a warning about this behavior to the docs.

Cheers,
Martin

### DIFF
--- a/Support/Help_Articles/Calendar/Calendar_Exporting_Data/README.md
+++ b/Support/Help_Articles/Calendar/Calendar_Exporting_Data/README.md
@@ -106,7 +106,7 @@ sqlite> select * from Components;
 sqlite> .quit 
 ```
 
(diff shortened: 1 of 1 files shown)


pvuorela:
Good notice but at the moment maybe also a bit early when we don't have such a version in any release. Once sqlite is updated, it would be good to note the OS release around here so people don't have to start figuring out which sqlite version they might have.

marhop:
Fair enough! I copied the database from phone to laptop, so I wasn't tied to the sqlite version shipped with Sailfish OS. So I guess this depends on the context in which you expect people to apply the instructions in the docs. I can well imagine that someone (like me) will backup the database on their laptop and export it to CSV with whatever version of sqlite is installed there (i.e., *not* on the phone); but if you wish to keep the focus strictly on Sailfish OS then I do agree that it may be too early to update the docs.

So please feel free to accept or close this PR at any time just as you see fit!

pvuorela:
Maybe I'd at this point explicitly mention 'Note that if you are using sqlite version 3.52+ on a computer...' to avoid people checking what sailfish has. Once sailfish version gets updated that can be then moved to the basic command set and invert the notice.

The '3.51 and earlier' I'd skip to keep this short. We don't need to explain it too much.

marhop:
Sounds like a good middle ground! I updated the PR.

pvuorela (approved):
Should be good now, thanks.

PR #467, opened 2026-05-26
Add TOH detection to documentation

Tomin1:
This describes the needed resistor and the memory chip but is missing exact values, memory chips types and schematics. Data layout is described in a high level but details are to be added. The data format is not yet set in stone and any feedback on that is very welcome.

I intend to add schematics, resistor values and memory chip types still before merging this. Perhaps also clarify that what the CBOR data should contain.

I also adjusted some of the earlier texts based on feedback and what I got from engineers working on the hardware.


Develop/Hardware/JP2601-TOH/README.md
```diff
@@ -50,3 +56,85 @@ Those are the main way of attaching TOHs.
...
+
+**I²C memory chip must have target device address of 0x50.**
+This is a very common memory chip address and also used for example for EDID info on displays.
+However different memory chips have different ways of addressing the data.
+To facilitate that some ID resistor values have been assigned to certain types of I²C chips.
+
+**Use 1% tolerance or better resistors for ID.**
+This allows for a good range of values and takes into account variance in ADC accuracy.
+
+<!-- TODO: ID resistor values and matching memory chip types to be added -->
```
Tomin1: I'm currently thinking that there would be at least a memory chip with one or more blocks of 256 bytes with 8-bit addressing and then also a memory chip with one or more blocks of 16,384 or more bytes with 16-bit addressing. These seem to be the most common types.

Are there any other types we should consider?

I expect that most mcus with emulated memory chips will use the latter type as it will have sufficient space with just one target address (0x50) and typically one I2C device on an mcu can only provide one target address. (Though e.g. on RP2040 one can connect both I2C devices on the same bus and also implement more I2C devices with PIO so multiple address can be used.)

Develop/Hardware/JP2601-TOH/README.md
```diff
@@ -50,3 +56,85 @@ Those are the main way of attaching TOHs.
...
+If TOH needs to control INT pin, then it should use a more complex circuit that pulls the pin low before there is power.
+The default is always to pull INT pin down but TOH can set the pin high once it has been given permission to use INT from the device and it needs to send an interrupt.
+
+When TOH is attached the device sees an interrupt on INT pin and checks ID pin voltage set by the resistor.
+Depending on the voltage the next step may be to turn on 5V output and read I²C memory chip at address 0x50.
+The memory chip content describes what TOH has been connected and how it should be handled.
+
+Detaching TOH is detected by INT pin becoming high and the device reading a voltage that tells there is no ID pin pull-down present any more.
+
+<!-- TODO: Insert some schematics for the circuits -->
```
Tomin1: I would add a minimal circuit for TOH (i.e. grounding INT, pull-down for ID, a regulator and a memory chip) and then possibly also some example of the 'more complex circuit' for pull-down.

Develop/Hardware/JP2601-TOH/README.md
```diff
@@ -50,3 +56,85 @@ Those are the main way of attaching TOHs.
...
+#### Payload data
+
+The CBOR data format that payload uses is a common JSON like binary format that has much lower overhead.
+The purpose of payload data is to define things like whether 5 V output should be left on after reading the memory chip,
+if the attached TOH is intended for example as a charging TOH or as an input TOH,
+what ambience should be applied if any,
+or a website for the vendor or product to get more information.
+Extensibility of CBOR allows to add new data in future TOHs.
+
+<!-- TODO: Add a description of the expected payload content, e.g. keys to be found in the data -->
```
Tomin1: The main point of the data is to allow Sailfish UI to show some information about TOHs and to react correctly to any attached TOH before TOH specific software has been installed. My current thinking is that we'd have a map that would have short string keys. For example:

- `SC` - schema version, integer, just put zero here for now
- `SN` - serial number, string
- `VN` - vendor name, string
- `PN` - product name, string
- `VS` - vendor website, url string
- `PS` - product website, url string
- `PO` - leave power out enabled after detection, boolean
- `PI` - is this power input TOH, boolean
- `AM` - ambience identifier, string
- `EX` - extra data for extending, another map for TOH specific content

I think everything here is optional for now but perhaps schema version should be always there if there is any content at all. I used two letter strings to save space but EEPROMs are quite cheap in larger sizes too so were are not too limited by that.

What else information would we like to have? Do you see anything that we should improve on this idea?

Develop/Hardware/JP2601-TOH/README.md
Tomin1: Okay, the 'more complex circuit' got reduced down to some resistors so nothing difficult there. But I would add the minimal circuit anyway.

Develop/Hardware/JP2601-TOH/README.md
```diff
@@ -50,3 +59,86 @@ Those are the main way of attaching TOHs.
...
+A TOH with a microcontroller unit may emulate a memory chip at address 0x50.
+In that case it should use a resistor that corresponds to the emulated memory chip.
+
+### Memory chip content
+
+The table below describes what the memory chip must contain.
+First value at address 0 in the order of the table.
+All integers longer than one byte are big-endian (aka network byte order).
+
+| Name         | Size     | Value |
```
Tomin1: This could probably use offset column as well.

Develop/Hardware/JP2601-TOH/README.md
Tomin1: One thing I'm wondering about here is that if we should separate those larger memory chips with 16,786 byte and 32,786 byte blocks to two separate groups, or since they both take 16-bit addresses should we just include in the data how large the pages are. There is some space (see that reserved column) where we could put 8-bit value that tells what power of two the block size is.

Or it could be just part of the byte if we want so, we can just assume some minimum size as I don't expect anything but 2^14, 2^15 and 2^16 to exist for chips 16-bit data addressing. It might make sense to start from 2^9 (i.e. the smallest block size that needs more than one byte to address) so that would be zero and eight values can be represented by 3 bits so we have 5 bits for something else if needed later. :slightly_smiling_face: (And for 8-bit addressing the most it can present is 256 bytes and probably every chip that span multiple blocks use that so I guess such value makes little sense.)

Anyway I suppose that doesn't really matter unless one puts more than one chip to the board because in general it seems that there no 16-bit addressing chips that have multiple target addresses unless they are really big and then they certainly use 64k blocks. And then one might ask if it actually makes sense to even support that large memory chips for this. Why would anyone need that much data in a TOH for the purpose of detecting one? :wink:

Tomin1: Also do you think if resistors for TOHs without memory chips would be useful? The are basically just two types that could make sense: one that does not enable power output and one that does enable it. Power input can use used regardless which is probably the main use case for TOH that does not enable power output. I think those TOH types should be discouraged but I do wonder if people would find these useful and if we should include them. Of course we can add them later on as well.

Tomin1: Now that I've looked into this a bit more, I think I'll add these types:

- 10 kohm resistor for 8-bit addressed memory chip with up to 256 byte pages
- 15 kohm resistor for 16-bit addressed memory chip with up to 65,536 byte pages

These are probably the ones that come up. If the data spans multiple blocks, then we assume maximum size for each block. We anyway read only up to the limit that the data told us to read so having a smaller page doesn't hurt. So just don't put two 16k chips to 0x50 and 0x51 and expect it to work. If we need to, we can add the block size to the start into the reserved bits so it can be read before we reach the smallest block size.

Now the question is mostly that whether we'd like to have those memory chip -less variants.

Develop/Hardware/JP2601-TOH/README.md
Tomin1: I added some of the keys. We can add more as we go.

direc85 (approved):

PR #468, opened 2026-06-13
Fix Sony developerworld links

nephros:
The website has moved.

I have not verified each new link, but the paths seem to have stayed roughly the same.


nephros:
Note that the links (esp. to the binaries) in the official install docs are also broken:

https://jolla.com/sailfishxinstall/)

pvuorela (approved):
Yea, seems to be redirecting there these days.

